PR #1, opened 2023-02-14
Aggregate data structure library

whitequark:
[Rendered](https://github.com/amaranth-lang/rfcs/blob/aggregate-data-structures/text/0000-aggregate-data-structures.md)


whitequark:
As this RFC predates the RFC process proper, the discussion is in amaranth-lang/amaranth#693; see that issue for additional historical context.

whitequark:
Per previous discussion, this RFC is **accepted**.

PR #2, opened 2023-02-16
Interface definition library

whitequark:
[Rendered](https://github.com/amaranth-lang/rfcs/blob/interfaces/text/0002-interfaces.md)


whitequark:
> How does this fit into Python's type system? Ideally it could be used for verification.

I agree. I'm still exploring options here, so I have no definite answer for you. I am definitely going to make sure that whatever solution we end up with *can* fit into Python's type system.

josuah:
The `__getitem__` binding greatly helps making the syntax more lightweight.

`Member(Flow.Out, unsinged(4))` becomes `In[unsigned(4)]`.

With the `[]` syntax, I picture `In` and `Out` being some mathematical set defined as all possible signatures, from which we would drag one in particular (they `[`contain`]` a signature).

Was it voluntary to avoid small wrapper classes?

```python
class In(Flow):
    def __init__(self, signature):
        super().__init__(self, Flow.In, signature):

class Out(Flow):
    def __init__(self, signature):
        super().__init__(self, Flow.Out, signature):
```

Then they look like some wrapper around a shape, adding a direction to them.

```python
input_port_template = Signature({
    "data": In(unsigned(4)),
    "vld":  In(Signal(1)),
    "rdy":  Out(Signal(1))
})
```

galibert:
Why is there both Interface and Signature?  Which is Interface adding?

josuah:
I understand `Signature`s a templates that can be used to create an Interface.

I think the use-case is to be able to define a Signature for commonplace internal ports like Wishbone, FIFOs, valid/ready, etc. and then plug multiple instances of these.

Multiple `Interface`s instantiated from the same `Signature` are likely compatible with each other.

whitequark:
Yep!

josuah:
A `Signature` is a bit like a "class", and an `Interface` is a bit like an object of that class.

Maybe it is possible to have signatures defined as classes (inheriting from a `Signature` class), and interfaces defined as objects of these?

Something a bit like how `data.Struct`s are made?

```
class WishbonePort(data.Interface):
    dat_r:  In(Signal(32))
    addr:   In(Signal(32))
    stb:    In(Signal(1))
    ack:    Out(Signal(1))
    ...
```

That looks good on paper, but:
Maybe that would require metaclasses.
Maybe that would be too cumbersome to use or implement.

whitequark:
> Maybe it is possible to have signatures defined as classes (inheriting from a `Signature` class), and interfaces defined as objects of these?
> 
> Something a bit like how [`data.Struct`](https://github.com/amaranth-lang/rfcs/blob/main/text/0001-aggregate-data-structures.md)s are made?

It's simpler. Any class that has a compatible interface matches a Signature. It doesn't have to be an `Interface`, it just needs to have the right fields.

whitequark:
Yeah

josuah:
I am trying to summarize a typical usage of that RFC:

```python
# defining the controller side in the signature by convention?
spi_signature = Signature({
    "cs":   Out[Signal(1)],
    "copi": Out[Signal(1)],
    "cipo": In[Signal(1)],
    "clk":  Out[Signal(1)],
})

class SpiControllerInterface(Interface):
    def __init__(self):
        super().__init__(spi_signature)

class SpiPeripheralInterface(Interface):
    def __init__(self):
        super().__init__(~spi_signature)
```

This looks rather minimal.

And if we add extra helpers:

```python
# defining the controller side in the signature by convention?
spi_signature = Signature({
    "cs":   Out[Signal(1)],
    "copi": Out[Signal(1)],
    "cipo": In[Signal(1)],
    "clk":  Out[Signal(1)],
})

class SpiControllerInterface(Interface):
    def __init__(self):
        super().__init__(spi_signature)

    # for simulation:

    def read_byte(self, data):
        pass

    def write_byte(self, data):
        pass

class SpiPeripheralInterface(Interface):
    def __init__(self):
        super().__init__(~spi_signature)

    def transfer_callback(self, callback):
        pass
```

Maybe it would be convenient to turn a Resource into a Signature (or rather, the other way around?) to share data formats between those.

galibert:
Any reason not to directly jump to/specify the later point?

josuah:
Here is my attempt at using Records in a way that looks like Interfaces/Signatures,
so that I can write new code ready for when they come:

```python
class SPISignature(Record):
    def __init__(self, *, ctrl, peri):
        return super().__init__([
            ("clk",     1,  ctrl), # ctrl and peri here are to
            ("copi",    1,  ctrl), # tell who drives the signal
            ("cipo",    1,  peri),
            ("cs",      1,  ctrl),
        ])

class SPICtrlInterface(SPISignature):
    def __init__(self):
        super().__init__(ctrl=DIR_FANOUT, peri=DIR_FANIN)

    def spi_ctrl_methods(self):
        pass

class SPIPeriInterface(SPISignature):
    def __init__(self):
        super().__init__(peri=DIR_FANOUT, ctrl=DIR_FANIN)

    def spi_peri_methods(self):
        pass
```

jfng:
* Would it be possible to expose other `Signal` parameters to the `Member` constructor ? I think `decoder` could be useful here.

* Will there be cases where `signature.freeze()` is implicitly called ? e.g. when `signature.__eq__(), or `Interface(signature)` are called.

* How would one attach metadata to an `Interface` ? The RFC mentions that `Signature` can be subclassed, is this an intended use case ?

text/0002-interfaces.md
```diff
@@ -0,0 +1,300 @@
...
+* For every port member with a given `path` and `first_shape` in `first`, there is a port member with the same `path` in each of `rest` with shape `rest_shape`, where `Shape.cast(first_shape) == Shape.cast(rest_shape)`, and there are no other members in any of `rest`. In other words, the width and signedness of all of port members must be equal, but the shape-castable objects specifying the width and signedness do not have to be.
+  * It is expected that any mismatch in signatures will be resolved (through wrappers, mutating signature members, or otherwise) before interfaces are being connected.
+* Either:
+  1. There is exactly one interface in `rest`, and for every pair of members `first_member` and `rest_member`, `first_member.flow == ~rest_member.flow`.
+  2. There is any amount of interfaces in `rest`, and for every pair of members `first_member` and `rest_member`, `first_member.flow == Out` and `rest_member.flow == In`.
+
+In both of the cases (1) and (2), the function, for each pair of an input port and an output of port with the same path, the function connects these as follows:
+```python
+m.d.comb += output_port.eq(input_port)
```
charlottia, line 269: I think the overloading of "out"/"output" and "in"/"input" in this illustrative code block has great potential to confuse — e.g. it's the _output stream_ of `SequenceSource` (which is described in the source with the member `Out[16]`) from the above example which would be the `input_port` here, and it's the _input stream_ of the `NumberSink` (described with `In[16]` in its signature) which is the `output_port`.

In the context of knowing only fan-out is supported (as noted later in § Future work), this makes sense, but in the context of case (2) saying "if you have multiple `rest` interfaces, all their members must be `Flow.In`", it may read as if there would be multiple `input_port`s and thus some kind of fan-in instead.

text/0002-interfaces.md
```diff
@@ -0,0 +1,300 @@
...
+## Drawbacks
+
+- The author considers `amaranth.lib.module` to be the best choice for this module, but it is ambiguous with `amaranth.hdl.dsl.Module`. In the author's view, if this ambiguity is a problem, then it is best to rename the latter (to `Builder`, since it is building up a netlist).
+
+
+## Alternatives and rationale
+
+- Do nothing. `Record` will continue to be used alongside the continued proliferation of ad-hoc implementations of similar functionality, and continue to impair the use of Amaranth components together.
+
+- Replace the `interface.lib.module.connect` free function with a function `Interface.connect` or a function `amaranth.hdl.dsl.Module.connect`.
```
charlottia, line 282:
```suggestion
- Replace the `amaranth.lib.module.connect` free function with a function `Interface.connect` or a function `amaranth.hdl.dsl.Module.connect`.
```

Also, this RFC no longer proposes an `Interface` class, and `Signature.connect` wouldn't work.

charlottia (commented):
I have some (very) minor comments from a read through of the changes since last week.

josuah:
Could you help neophytes like me to understand the nuance between the use of the new "forward"

```python
component.connect(m, component.forward(self.bus), self.inner.bus)
```

And merely assigning the child's object to the parent with python?

```python
self.bus = self.inner.bus
```

Then other modules trying to call `Outter.bus` would send data directly to the inner module.
This means no extra signal created on `.vcd` dumps. Is it bad?

charlottia:
@josuah I can think of a few reasons why one might not prefer doing it that way:

* For starters, it does confuse the hierarchy a bit and like you say, might make `vcd`s (and debug output from other stages of tooling, names in RTLIL/Verilog/etc.) less obvious. Connections made from whatever includes `Outer` to its `bus` will appear instead to be made directly to `Inner`'s `bus`.
* Maintainability over time might be affected too; in the context of `Outer`, it's not necessarily clear that `self.bus` is actually lifted straight from its inner object.
  * Down the track, someone might be tempted to do other things to `self.bus` without realising it could affect `Inner`'s operation.
    * This kind of thing can happen easily enough over enough time (and with enough layers).
* `forward`+`connect` happens in `elaborate()` since it's logic-based. If we did this assignment in `elaborate()`, someone else might've already taken a reference to the `Outer`'s `bus`, which the `Outer`'s logic would then never actually use. This means it has to happen in the constructor. 
  * Let's say we do so. This means we overwrite the `Outer`'s `bus` with the `Inner`'s `bus`.
  * Imagine someone later on modifies `Inner`'s `Signature`, changing its `bus` member such that it's now incompatible with how `Outer` defines its `bus` member.
    * With `forward`+`connect`, this would generate a diagnostic at the point of that attempted connection.
    * If we allow overwriting the attribute entirely this way, `Outer`'s actual interface no longer matches its signature. Any signature mismatch from a `connect` to `my_outer.bus` would refer to `Inner`'s signature for `bus` (!), but the worse result is simply things quietly being wrong.
      * e.g. when you thought `my_outer.bus.addr` was a `Signal(8)` and used it accordingly because that's what `Outer`'s signature says it should be, but now it's actually a `Signal(12)`—some of your logic assumed it would wrap at 256, and now it doesn't.
      * We could only allow this overwriting if the signatures are compatible, but it feels like a solution in search of a problem.
  * I think that means we should just not allow it (or at least support it) — otherwise many properties implied by having defined interfaces to begin with are weakened.
    * Just imo: not only should the identity of the attributes created by applying the signature be constant (i.e. no setattr/delattr), likewise the shape and reset. Otherwise we compromise conformance guarantees despite carrying/applying a given `signature` classvar. 

Also, maybe it's just me, but I've seen optimisations _improve_ when I make forwarding connections compared to doing this kind of thing. In a sense, you're just giving the toolchain more to work with (e.g. better preserving locality information), whereas the direct assignment is a kind of premature optimisation that doesn't have an equivalent in Verilog, and so may not be as easily anticipated.

whitequark:
> Could you help neophytes like me to understand the nuance between the use of the new "forward"
> 
> ```python
> component.connect(m, component.forward(self.bus), self.inner.bus)
> ```
> 
> And merely assigning the child's object to the parent with python?
> 
> ```python
> self.bus = self.inner.bus
> ```

You *could* do the latter. However, as @charlottia points out, there is a number of possible issues. Nevertheless, `self.bus = self.inner.bus` would technically work, it's just not what I would recommend in the documentation.

> Also, maybe it's just me, but I've seen optimisations _improve_ when I make forwarding connections compared to doing this kind of thing.

After a planned toolchain improvement later this year these two cases should result in topologically equivalent HDL being emitted, so that's probably not going to be any concern after that.

whitequark:
This RFC is no longer a draft! It will remain open pending a prototype implementation, however.

whitequark:
The RFC is now in a state I consider ready for discussion.

charlottia:
Will accumulate results from testing the [implementation](https://github.com/whitequark/amaranth/compare/rfc-2-implement).

* The default definition of `Component.signature`, which implements deriving the signature from variable annotations, only looks at the annotations of the type of the object it was called on. e.g. if you define `class A(Component)` with its own annotations, `class B(A)` no longer derives those members, since they are not present in `B.__annotations__`.
  * I think we should follow the MRO chain up to `Component`, which requires a slight modification to the RFC:
  
    > `Component.signature` collects PEP 526 variable annotations in the class<ins>'s method resolution order chain up to `Component`</ins>, if any, [...]

    My wording here is a little awkward. The corresponding implementation:

    ```diff
    --- a/amaranth/lib/wiring.py
    +++ b/amaranth/lib/wiring.py
    @@ -716,9 +716,11 @@ class Component(Elaboratable):
     
         @property
         def signature(self):
    +        mro = type(self).mro()
             member_annotations = {
                 name: annot
    -            for name, annot in getattr(type(self), '__annotations__', {}).items()
    +            for subclass in mro[:mro.index(Component)]
    +            for name, annot in getattr(subclass, '__annotations__', {}).items()
                 if type(annot) is Member
             }
             if not member_annotations:
    ```

  * The alternative is to bless the current behavior. It is probably surprising, though not entirely without use — maybe you _do_ want to not have some of the superclass's members. That case is likely better handled by explicitly setting `signature` in the subclass.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-08-21#1692637574-1692638527;) this RFC on the 2023-08-21 weekly meeting. The disposition was to **merge**.

PR #4, opened 2023-02-21
Const-castable expressions

whitequark:
[Rendered](https://github.com/amaranth-lang/rfcs/blob/const-castable/text/0000-const-castable-exprs.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-02-27#1677517838-1677518403;) this RFC on the 2023-02-27 meeting. The consensus was to **merge**.

PR #5, opened 2023-02-21
Remove `Const.normalize`

whitequark:
[Rendered](https://github.com/amaranth-lang/rfcs/blob/remove-const-normalize/text/0000-remove-const-normalize.md)


josuah:
As I understood it, there is a drop-in replacement?
Then anyone can do a textual conversion from `Const.normalize(%1, %2)` to `Const(%1, %2).value` in source?
These would make breakage more acceptable as trivial to fix.

whitequark:
> Then anyone can do a textual conversion from `Const.normalize(%1, %2)` to `Const(%1, %2).value` in source?

Yeah.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-02-27#1677517308-1677517836;) this RFC on the 2023-02-07 meeting. The consensus was to **merge**.

PR #7, opened 2023-02-27
Const-castable Amaranth value to Python value conversions

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/const-conversions/text/0000-const-conversions.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-03-06#1678126340-1678127830;) this RFC n the 2023-03-06 weekly meeting, and concluded that more discussion is necessary.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-05-22#1684777810-1684778504;) this RFC on the 2023-05-22 weekly meeting. The disposition was to **close** as the motivational section is no longer impossible otherwise and the drawbacks are too great.

PR #8, opened 2023-03-03
Aggregate data structure extensibility

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/aggregate-extensibility/text/0000-aggregate-extensibility.md)


whitequark:
There is an example implementation in https://github.com/whitequark/amaranth/commits/rfc-8.

PR #9, opened 2023-03-03
Constant initialization for shape-castable objects

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/aggregate-const-init/text/0000-const-init-shape-castable.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-03-06#1678123847-1678125300;) this on the 2023-03-06 weekly meeting. The disposition was to **merge**.

PR #11, opened 2023-03-05
Minimal type annotation support

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/minimal-type-hints/text/0000-minimal-type-hints.md)

### DIFF
--- a/text/0000-minimal-type-hints.md
+++ b/text/0000-minimal-type-hints.md
@@ -1,0 +1,139 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [amaranth-lang/rfcs#0000](https://github.com/amaranth-lang/rfcs/pull/0000)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# Minimal type annotation support
+
+## Summary
+[summary]: #summary
+
+Add the minimal amount of functionality required for Amaranth developers to be able to use type hints in their code that are recognized by IDEs such as Visual Studio Code.
+
+## Motivation
+[motivation]: #motivation
+
+[PEP 484] type hints have been steadily gaining popularity and there's been numerous requests to add some level of support for PEP 484 in Amaranth. There are some difficulties in adding these to Amaranth code, which is why Amaranth isn't currently typed; discussing this is out of scope for this RFC.
+
+[RFC 1] adds functionality to Amaranth itself relying on the type hint syntax. The example given in the RFC is:
+
+```python
+from amaranth import *
+from amaranth.lib import data
+
+
+class IEEE754Single(data.Struct):
+    mantissa: 23
+    exponent: 8
+    sign:     1
+```
+
+Unfortunately, `23` is not a valid Python type hint, nor is the more explicit form of `unsigned(23)`. While this definition will result in IDEs picking up the *existence* of the fields, they will not show any completion for `IEEE754Single().mantissa.<cursor>`.
+
+The following definition, while somewhat uglier, does provide full IDE autocompletion functionality:
+
+```python
+class IEEE754Single(data.Struct):
+    mantissa: Value[23]
+    exponent: Value[8]
+    sign:     Value[1]
+```
+
+[PEP 484]: https://peps.python.org/pep-0484/
+[RFC 1]: 0001-aggregate-data-structures.md
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The `amaranth.hdl.Value` class becomes [subscriptable]; calling `Value[s]`, where `s` must be shape-castable, returns a [typing.GenericAlias] object with `s` as its sole argument. This object can be used as usual in type hints, and in addition it is accepted by `amaranth.lib.data.Struct` and `amaranth.lib.data.Union`.
+
+[subscriptable]: https://docs.python.org/3/library/stdtypes.html#types-genericalias
+[typing.GenericAlias]: https://docs.python.org/3/library/types.html#types.GenericAlias
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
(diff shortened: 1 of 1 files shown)


tilk:
You could be interested in the fact that the Coreblocks project [uses type stubs for Amaranth](https://github.com/kuznia-rdzeni/coreblocks/tree/master/stubs/amaranth). We use pyright for type checking, as mypy was found to be not good enough for the task.

whitequark:
Thanks, that is useful information.

stafverhaegen-chipflow:
Originally the type system was architected to not have run time effects, so one should be able to remove the type hints and the code should have the same effect. [RFC 1] is using the type syntax and not following that rule:

```python
class IEEE754Single(data.Struct):
    mantissa: Value[23]
    exponent: Value[8]
    sign:     Value[1]
```

is not the same as:

```python
class IEEE754Single(data.Struct):
    mantissa
    exponent
    sign
```
 the latter is not even valid python code.

I'm thinking of changing syntac to something as follows but this needs change to [RFC 1]:
```python
class IEEE754SingleLayout(data.StructLayout):
    mantissa = unsigned(23)
    exponent = unsigned(8)
    sign = unsigned(1)

class IEEE754Single(data.Struct):
    layout = IEEE754SingleLayout()
```
For people who want type annotations the latter class could then also be defined as:
```python
class IEEE754Single(data.Struct):
    layout = IEEE754SingleLayout()

    mantissa: Value[23]
    exponent: Value[8]
    sign: Value[1]
```

whitequark:
> [RFC 1] is using the type syntax and not following that rule:

This is a topic of ongoing discussion in the Python community: whether the annotation syntax is the type syntax, or not. [PEP 563](https://peps.python.org/pep-0563/) and [PEP 649](https://peps.python.org/pep-0649/) are both indefinitely deferred because [there is no consensus](https://lwn.net/Articles/866613/). Guido's position is that annotations are only for types, whereas other important voices consider them a general purpose mechanism.

Currently, Amaranth is using annotations as a general purpose mechanism. It seems like it won't be feasible to also use them as type syntax because the changes required to adapt them to be valid types are too onerous as there was significant pushback against that change on the meeting last week.

> For people who want type annotations the latter class could then also be defined as:

`Value[23]` isn't a valid type. `Value[U23]` where `U23` is defined somewhere else would be.

> I'm thinking of changing syntac to something as follows but this needs change to [RFC 1]:
> ```python
> class IEEE754SingleLayout(data.StructLayout):
>     mantissa = unsigned(23)
> ```

If we extend `Signal` such that the following works:

```python
obj = Signal(IEEE754SingleLayout, reset={"mantissa": 1})
m.d.comb += obj.mantissa.eq(2)
```

then we won't need to have `data.Struct` and `data.Union` at all, sidestepping the issue with type annotations.

However, I'm not sure if people will get IDE autocompletion then, which was the primary motivation for this change.

stafverhaegen-chipflow:
> `Value[23]` isn't a valid type. `Value[U23]` where `U23` is defined somewhere else would be.
> 
So `Literal[23]` is then a special case ?

whitequark:
> So `Literal[23]` is then a special case ?

I think so yeah

stafverhaegen-chipflow:
> However, I'm not sure if people will get IDE autocompletion then, which was the primary motivation for this change.

For me, in the long term I would like the type support to flag something like the following as an error:
```python
sig: Value[8] = Signal(10)
```
For autocompletion just `Value` as type would be enough:
```python
class IEEE754Single(data.Struct):
    layout = IEEE754SingleLayout()

    mantissa: Value
    exponent: Value
    sign: Value
```

whitequark:
> For autocompletion just `Value` as type would be enough:

I think this is needlessly repetitive in a way I'd like to avoid. I would hesitate suggesting this in our docs.

> For me, in the long term I would like the type support to flag something like the following as an error:

I think this isn't actually viable because Python type systems aren't flexible enough to describe Amaranth patterns like variable width signals in FIFO interfaces.

stafverhaegen-chipflow:
> > So `Literal[23]` is then a special case ?
> 
> I think so yeah

As you mentioned in the RFC isn't `types.GenericAlias` from python 3.9 the way to make it not a special case ?

BTW, if choice is between minimum python 3.9 version and mypy/pytype support and older python version support and only pyright/pylance I prefer the former.

whitequark:
> As you mentioned in the RFC isn't `types.GenericAlias` from python 3.9 the way to make it not a special case ?

That doesn't actually usefully interact with type checking, it's just a runtime container I picked.

---

In general, the more I'm looking into this, the more pessimistic I am about the usefulness of Python type systems as applied to Amaranth.

tilk:
As a maintainer of a project which uses Amaranth with Pyright (manually stubbed), I'd like to state we're unhappy with this decision - this essentially makes defining `Struct`s as classes unusable for our project. I hope this will be revisited later.

whitequark:
@tilk This decision can be revisited through the RFC process. If you have a solution to the issues we uncovered while working on this RFC I would encourage you to draft and submit your own.

whitequark:
@tilk I should also mention that if you are a heavy user of the language, you should participate in the weekly meetings (1700 UTC every Monday, on IRC/Matrix channel #amaranth-lang on libera.chat) since that gives you an opportunity to direct the language's evolution.

tilk:
@whitequark Thank you for the invitation. Unfortunately I am busy (lectures) during this time, so I won't be able to join until July.

whitequark:
@tilk Which part of the globe are you in and what timing would work for you? I've considered shifting it in time already, so I would be able to take your needs into account as well if I do.

tilk:
@whitequark I'm using Central European Time, CEST currently; I just sometimes have classes with students on late hours, and often schedule other activities at such hours. At the moment, I have the most time available on Thursdays.

whitequark:
> It should be possible to use annotated types to have both actual type and shape information:

This is something that would require another RFC. I am not currently pursuing any more work involving Python typecheckers myself.

PR #12, opened 2023-04-12
Better definition of clock domains

galibert:
[Rendered](https://github.com/galibert/amaranth-rfcs/blob/clock-domains/text/0012-better-clock-domains.md)

### DIFF
--- a/text/0000-better-clock-domains.md
+++ b/text/0000-better-clock-domains.md
@@ -1,0 +1,357 @@
+- Start Date: 2023-04-10
+- RFC PR: [amaranth-lang/rfcs#0000](https://github.com/amaranth-lang/rfcs/pull/0000)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# Better clock domains
+
+## Summary
+[summary]: #summary
+
+(Re)define what clock domains exactly are and what one can do with
+them, making them more explicit and flexible for a better abstraction
+of clocks in the designs.
+
+## Motivation
+[motivation]: #motivation
+
+Clock domains in amaranth are supposed to be first-level concepts,
+except they're not entirely at this point.  Reset is a little tacked
+on, and enable has not even entirely managed to be tacked.  OTOH they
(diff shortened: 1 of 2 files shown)


josuah:
The new syntax, in particular the new ClockDomain constructor, makes it a whole lot easier to read!

To have an insight on who used the existing API:
https://github.com/search?q=enableinserter+language%3APython&type=code&l=Python

whitequark:
This all is broadly reasonable and in line with what I've been planning already. It is a very large change however, and I think we won't get around to it particularly soon. Ultimately something like this needs to be done before 1.0, and probably this year.

jevinskie:
Has anybody thought about adding clock frequency information to ClockDomains?

I opened https://github.com/enjoy-digital/litex/pull/1310 because I wanted build-time assertions to ensure that a hard IP block wasn’t being overclocked.

Perhaps `freq`, `freq_min`, `freq_max` and maybe a `freq_margin` (for PLL outputs that have guaranteed tolerances) could be useful? Or would that be better served by a new RFC?

whitequark:
> Has anybody thought about adding clock frequency information to ClockDomains?

Yes. This is a rather complex problem in situations where frequency can change at runtime. It should not be a part of this RFC.

whitequark:
Reopened as requested by the author.

whitequark:
I've listed the challenges associated with the current clock domain system in [this conversation](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-01#1709318917-1709327968;). To move this proposal forward, it will have to give an answer to these challenges, especially the interaction of sync reset, async reset, and enable.

More broadly, I think that the domain scoping and the domain control set semantics are topics that can probably be addressed separately. Given how troublesome domain scoping is for both understanding the language and for implementing it (NIR and the platform layer struggle with it) this may be worth addressing separately.

galibert:
That rfc is around a year and half old and has never been discussed.  By now, I'm not close enough to top-of-tree development of amaranth to be able to see whether it still makes sense, and I'm starting to have a feeling its existence is preventing other people to propose things on the topic.  So it's time to close it.

PR #15, opened 2023-05-12
Lifting shape-castable objects

whitequark:


whitequark:
An implementation of this RFC is available in https://github.com/amaranth-lang/amaranth/pull/780.

whitequark:
> Code like this stops working and it's not clear how to replace it:
> 
> ```python
> class RGBLayout(data.View):
>     def __init__(self, target=None, *, r_bits, g_bits, b_bits, **kwargs):
>         super().__init__(data.StructLayout({
>             "red":   unsigned(r_bits),
>             "green": unsigned(g_bits),
>             "blue":  unsigned(b_bits)
>         }, target, **kwargs))
> 
>     def brightness(self):
>         return (self.red + self.green + self.blue)[-8:]
> ```

This was a piece of code from the doc I just wrote and the answer is that it doesn't work in first place, at all. I rewrote the example to be:

```python
class RGBPixelLayout(data.StructLayout):
    def __init__(self, r_bits, g_bits, b_bits):
        super().__init__({
            "red":   unsigned(r_bits),
            "green": unsigned(g_bits),
            "blue":  unsigned(b_bits)
        })

    def __call__(self, value):
        return RGBPixelView(self, value)

class RGBPixelView(data.View):
    def brightness(self):
        return (self.red + self.green + self.blue)[-8:]
```

and this will work just fine with the proposed RFC. (It is also nicer in first place!)

whitequark:
We've [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-05-15#1684171125-1684172126) this RFC on the 2023-05-15 weekly meeting. The disposition was to **merge**.

PR #16, opened 2023-06-26
A CSR register API for amaranth-soc

jfng:
[Rendered](https://github.com/jfng/amaranth-rfcs/blob/soc-csr/text/0016-soc-csr-regs.md)

### DIFF
--- a/text/0016-soc-csr-regs.md
+++ b/text/0016-soc-csr-regs.md
@@ -547,18 +547,13 @@ The `csr.reg.Builder` class can build a `MemoryMap` from a group of `csr.Registe
     * upon entry, creates a scope where registers added by `self.add()` are assigned to a cluster named `name`;
     * raises a `ValueError` if `self` is frozen;
     * raises a `TypeError` if `name` is not a non-empty string;
-    * raises a `TypeError` if `size` is not a positive integer;
-    * raises a `TypeError` if `offset` is neither a positive integer or 0;
-    * raises a `ValueError` if `offset` is not word-aligned (i.e. a multiple of `self.data_width // self.granularity`);
+    * raises a `ValueError` if `name` is already assigned to another register or `Cluster`;
 
 - a `.Index(self, index)` context manager method, which:
     * upon entry, creates a scope where registers added by `self.add()` are assigned to an array index `index`;
     * raises a `ValueError` if `self` is frozen;
     * raises a `TypeError` if `index` is neither a positive integer or 0;
     * raises a `ValueError` if `index` is already assigned to another `Index`;
-    * raises a `TypeError` if `size` is not a positive integer;
-    * raises a `TypeError` if `offset` is neither a positive integer or 0;
-    * raises a `ValueError` if `offset` is not word-aligned (i.e. a multiple of `self.data_width // self.granularity`);
 
 - a `.as_memory_map(self)` method, that converts `self` into a `MemoryMap`. `self.freeze()` is implicitly called as a side-effect.
 
--- a/text/0016-soc-csr-regs.md
+++ b/text/0016-soc-csr-regs.md
@@ -1,0 +1,645 @@
+- Start Date: 2024-02-02
+- RFC PR: [amaranth-lang/rfcs#16](https://github.com/amaranth-lang/rfcs/pull/16)
+- Amaranth SoC Issue: [amaranth-lang/amaranth-soc#68](https://github.com/amaranth-lang/amaranth-soc/issues/68)
+
+# CSR register definition RFC
+
+## Summary
+[summary]: #summary
+
+Add primitives to define CSR registers.
+
+## Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


whitequark:
This looks like an RFC (maybe RFC draft) to me, not a pre-RFC!

(In general I've seen people go too far with sometimes even multiple pre- suffixes if they don't feel absolutely confident about the document, which is a bit silly--the RFC stands for Request For Comments after all! We want the comments because we know the document is imperfect, there's no shame in calling it an RFC.)

jfng:
The TODOs for some key parts of the RFC (register banks, prior art) made me think so. It is in a state where comments are welcome, but the proposal itself is incomplete.

I'll just remove the prefix.

whitequark:
Ah! I call this an "RFC draft". I think of "pre-RFC" as a document that, *even when complete* does not form a viable proposal for inclusion of a feature in the language, but rather only a starting point for a proper RFC.

jfng:
This RFC was discussed during the [2023-12-01](https://libera.irclog.whitequark.org/amaranth-lang/2023-12-01#1701450082-1701459939;) weekly SoC meeting. The discussions centered around an issue with `csr.Field`:

@tpwrules:
- `csr.Field` instances created as variable annotations of a `csr.Register` are global state; they will cause driver conflicts if the latter is instantiated more than once.

@galibert, @whitequark (in a broader scope than this RFC alone):
- Considering a `Signature` as the "type" of a `Component` is confusing. Some components use identical signatures for all their instances, while others use parametric signatures that differ for each instance.

@whitequark:
- When using variable annotations, the signature of a `Component` should be a class attribute, created and frozen in `__init_subclass__`.
- `csr.Field` should be a plain object (similarly to `lib.wiring.In`), instead of a `Component`:
  * an example would be `foo: csr.Field(csr.field.RW, unsigned(1), reset=1)`;
  * `csr.field.RW` (and others) would still be a `Component`.

jfng:
Discussed during the [2024-01-12 SoC]( https://libera.irclog.whitequark.org/amaranth-lang/2024-01-12) meeting, with the following suggested:

- add a `FieldAction` base class for field action components;
- rename `FieldMap` and `FieldArray` to `FieldActionMap` and `FieldActionArray`.

These changes should help clear confusion between field definitions and instances throughout the RFC.

jfng:
Discussed during the [2024-01-19](https://libera.irclog.whitequark.org/amaranth-lang/2024-01-19#1705683597-1705687826;) SoC meeting:

(@tpwrules) `csr.action.R` and `csr.action.W` should expose `r_stb` and `w_stb` strobes, respectively:
* peripherals will sample strobes to trigger side-effects, which are use-cases for `action.R`/`action.W` field actions.
* the `FieldPort` interface is specific to the `FieldAction`/`Register` connection, and its signals shouldn't be sampled by peripherals.

(@whitequark) Support for arrays of registers will be needed eventually, as they are part of the SVD specification. They may come in a later RFC, but the current proposal prevents them from being implemented, which is a problem.

jfng:
> Discussed during the [2024-01-19](https://libera.irclog.whitequark.org/amaranth-lang/2024-01-19#1705683597-1705687826;) SoC meeting:
> 
> (@tpwrules) `csr.action.R` and `csr.action.W` should expose `r_stb` and `w_stb` strobes, respectively:
> 
>     * peripherals will sample strobes to trigger side-effects, which are use-cases for `action.R`/`action.W` field actions.
> 
>     * the `FieldPort` interface is specific to the `FieldAction`/`Register` connection, and its signals shouldn't be sampled by peripherals.
> 
> 
> (@whitequark) Support for arrays of registers will be needed eventually, as they are part of the SVD specification. They may come in a later RFC, but the current proposal prevents them from being implemented, which is a problem.

Updated to address the above feedback.

jfng:
Discussed during the [2024-02-02](https://libera.irclog.whitequark.org/amaranth-lang/2024-02-02#1706893261-1706897378;) SoC meeting:

(@whitequark) Introspection methods in `csr.RegisterMap` (e.g. `.flatten()`) can be misused (e.g. before a `RegisterMap` is frozen) and aren't useful right now, outside of unit tests. They shouldn't be part of the API, and remain an implementation detail instead.

(@Fatsie, @whitequark) `csr.RegisterMap` is a misleading name, as it does not have `MemoryMap`-like functionality. `csr.Builder` is less connotated.

The disposition was to **merge**, with the above changes.

whitequark:
@jfng Let me know when you've made the changes.

jfng:
> @jfng Let me know when you've made the changes.

Done. I also squashed previous commits, as they didn't follow the `"RFC #<num>: <msg>"` convention.

whitequark:
Should I merge this now?

jfng:
In ~1 hours, I'm giving it one last read.

jfng:
You can merge it now.

whitequark:
Congratulations on the first Amaranth SoC RFC being merged! :tada:

Here's for hundreds more.

PR #17, opened 2023-08-07
RFC: remove the `log2_int(n, need_pow2=True)` function.

jfng:
[Rendered](https://github.com/jfng/amaranth-rfcs/blob/remove-log2-int/text/0017-remove-log2-int.md).

### DIFF
--- a/text/0000-remove-log2-int.md
+++ b/text/0000-remove-log2-int.md
@@ -7,7 +7,7 @@
 ## Summary
 [summary]: #summary
 
-Remove the `log2_int(n, need_pow2=True)` function.
+Replace `log2_int` with two functions: `ceil_log2` and `exact_log2`.
 
 ## Motivation
 [motivation]: #motivation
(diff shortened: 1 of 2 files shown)


mwkmwkmwk:
The `log2_int` function, while badly named and with questionable behavior on negative / zero inputs, is a necessity in digital logic.  The suggested replacement using `math.log2` is subtly and dangerously wrong, being float-based:

```
>>> log2_int((1 << 64) + 1, need_pow2=False)
65
>>> math.ceil(math.log2((1 << 64) + 1))
64
```

I believe this RFC is a non-starter without providing better replacement functions in `amaranth` (perhaps `ceil_log2` and `exact_log2`)?

whitequark:
I agree with @mwkmwkmwk.

jfng:
Updated to address feedback from the 2023-08-7 weekly meeting.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-01-08#1704734248-1704734942;) this RFC on the 2024-01-08 weekly meeting. The disposition was to **merge**, with the change that `ceil_log2(0) == 0`.

whitequark:
Thank you!

PR #25, opened 2023-08-31
Amend RFC #2 to add `Signature.flatten`

whitequark:
This function isn't easy to implement with the existing API and is quite useful when connecting `lib.wiring` based interfaces to the outside world or components in other HDLs.


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-09-04#1693846917-1693847392;) this amendment on the 2023-09-04 weekly meeting and the disposition was to **merge**.

PR #26, opened 2023-09-03
RFC for Strobe Generator

galibert:
[Rendered](https://github.com/galibert/amaranth-rfcs/blob/strobegen/text/0026-strobegen.md)

### DIFF
--- a/text/0000-strobegen.md
+++ b/text/0000-strobegen.md
@@ -1,0 +1,130 @@
+- Start Date: 2023-09-03
+- RFC PR: [amaranth-lang/rfcs#0000](https://github.com/amaranth-lang/rfcs/pull/0000)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# Strobe Generator
+
+## Summary
+[summary]: #summary
+
+Introduce components for generating enable streams for clock division.
+
+## Motivation
+[motivation]: #motivation
+
+FPGAs tend to be way happier when the design uses a very low number of
+real clock signals, and various clocks are in fact a combination of a
+base, high-frequency clock and a regular enable signal that virtually
+divides it.  In particular, that avoid all problems linked to
+clock-domain-crossing.
+
+This RFC introduces in amaranth.lib three components that generate
+such an enable stream for any strobe frequency.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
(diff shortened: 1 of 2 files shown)


whitequark:
@galibert It is conventional to add a "rendered" link in the description of your RFC. You can get it by going to your branch and finding the file on GitHub: https://github.com/galibert/amaranth-rfcs/blob/strobegen/text/0026-strobegen.md

whitequark:
My personal view on this is that I don't think patterns that take less than 5 lines of Amaranth code normally should be added in the standard library. You can just write them out in your code.

In fact I'd like to see these removed from the standard library in the cases where they are added, e.g. the accepted RFC 18: https://amaranth-lang.org/rfcs/0019-remove-scheduler.html

galibert:
It's five lines of code, and 70 lines of explanation for why it works.  [This](https://github.com/MiSTer-devel/AtariST_MiSTer/blob/master/rtl/mfp/CEGen.vhd) is the kind of reason why I think best practices have a place in the lib.

whitequark:
Reopened as requested by the author.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-01-08#1704735079-1704736749;) this RFC on the 2024-01-08 weekly meeting. We did not reach a consensus.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-02-19#1708363271-1708365234) this RFC on the 2024-02-19 weekly meeting. The disposition was to **close**, with one abstention (@Fatsie). The following reasoning was provided:

@whitequark:

> my vote on RFC 26 is: close, for the following reasons:
>
> - it is not difficult to implement this functionality in-place and in my view this will usually be worth it because it will be integrated with some wider reset or clocking scheme
> - having three new classes for this purpose does not have a high enough utility for the increase in our API surface
> - the functionality could be implemented in an external library and later pulled in if we find that it was widely useful
>
> and the following note:
> - if we add EnableSignal and platform.unsafe_add_multicycle_constraint, we can add something like FixedStrobeGenerator as it will provide a clear safety benefit

In addition we have noted that:
- discoverability of Amaranth libraries should be improved (@galibert, @cr1901)
- `lib.coding` would not have made it to the standard library if it was proposed as an RFC and should probably be removed

Thank you for this proposal, @galibert.

FatsieFS:
BTW, it was @jfng who abstained; I voted for closure.

whitequark:
I misread the message. Thanks, I've updated the resolution.

PR #28, opened 2023-10-05
Add RFC to allow overriding Value operators

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/override-value-operators/text/0028-override-value-operators.md)


whitequark:
Thank you for starting this work! This is something I would eventually write if no one else did.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-10-30#1698685666-1698688900) this RFC on the 2023-10-30 weekly meeting. The disposition is to **merge**.

text/0028-override-value-operators.md
```diff
@@ -0,0 +1,65 @@
+- Start Date: 2023-10-05
```
whitequark, line 1: This should be today's date (of the meeting where the RFC was accepted), not the date you drafted it.

PR #29, opened 2023-10-06
RFC #9: clarify that `.const()` can return a value-castable.

zyp:


text/0009-const-init-shape-castable.md
```diff
@@ -44,9 +46,9 @@ A method `def const(self, obj):` is added on `ShapeCastable`.
 
 The `Signal(shape, reset=)` constructor is changed so that if `isinstance(shape, ShapeCastable)`, then `shape.const(reset)` is used instead of `reset`.
 
-The `.const()` instance method is implemented on `Layout` to accept a `Sequence` or `Mapping` instance and returns a `Const` with a bit pattern that has the fields set to the given values. Overlapping fields are written in the order of iteration of the input. If the field shape is a shape-castable object, then the value for that field is computed using `Const.cast(value, field.shape)`.
+The `.const()` instance method is implemented on `Layout` to accept a `Sequence` or `Mapping` instance and returns a `View` over a `Const` with a bit pattern that has the fields set to the given values. Overlapping fields are written in the order of iteration of the input. If the field shape is a shape-castable object, then the value for that field is computed using `Const.cast(value, field.shape)`.
```
whitequark, line 49: There is no such thing as a `View` in general; it is not something that exists outside of `lib.data`, and therefore cannot be a part of this amendment.

whitequark (changes requested):
See inline comment.

text/0009-const-init-shape-castable.md
zyp: `Layout` is also part of `lib.data` so I figure this point is `lib.data`-specific.

whitequark: Oh, no, you are absolutely right. Sorry, I'm extremely tired.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-10-09#1696871314-1696870948;) this on the 2023-10-09 weekly meeting. The disposition was to **merge**.

PR #27, opened 2023-10-02
Testbench functions for the simulator

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/simulator-testbenches/text/0000-simulator-testbenches.md)

PR #36, opened 2023-12-11
Add an RFC for async testbench functions.

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/async-testbench-functions/text/0036-async-testbench-functions.md)

### DIFF
--- a/text/0036-async-testbench-functions.md
+++ b/text/0036-async-testbench-functions.md
@@ -18,7 +18,7 @@ A more expressive way to specify trigger/wait conditions allows the condition ch
 
 Passing a simulator context to the testbench function provides a convenient place to gather all simulator operations.
 
-Having `.get()` and `.set()` methods provides a convenient way for value castables to implement these in a type-specific manner.
+~~Having `.get()` and `.set()` methods provides a convenient way for value castables to implement these in a type-specific manner.~~
 
 ## Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
@@ -29,25 +29,27 @@ On the interface class, we can then implement `.send()` and `.recv()` methods li
 ```python
 class StreamInterface(PureInterface):
     async def recv(self, sim):
-        await self.ready.set(1)
+        await sim.set(self.ready, 1)
         await sim.tick().until(self.valid)
 
-        value = await self.data.get()
+        value = await sim.get(self.data)
 
         await sim.tick()
-        await self.ready.set(0)
+        await sim.set(self.ready, 0)
 
         return value
 
     async def send(self, sim, value):
-        await self.data.set(value)
+        await sim.set(self.data, value)
 
-        await self.valid.set(1)
+        await sim.set(self.valid, 1)
         await sim.tick().until(self.ready)
 
         await sim.tick()
-        await self.valid.set(0)
+        await sim.set(self.valid, 0)
 ```
+
+`await sim.get()` and `await sim.set()` replaces the existing operations `yield signal` and `yield signal.eq()` respectively.
 
 `sim.tick()` replaces the existing `Tick()`. It returns a trigger object that either can be awaited directly, or made conditional through `.until()`.
 
@@ -77,7 +79,7 @@ async def testbench(sim):
     await test_rgb(sim, 255, 255, 255)
 ```
 
-Since `.send()` and `.recv()` invokes `.get()` and `.set()` that a value castable (here `data.View`) can implement in a suitable manner, it is general enough to work for streams with arbitrary shapes.
+Since `.send()` and `.recv()` invokes `sim.get()` and `sim.set()` that in turn will invoke the appropriate value conversions for a value castable (here `data.View`), it is general enough to work for streams with arbitrary shapes.
 
 `Tick()` and `Delay()` are replaced by `sim.tick()` and `sim.delay()` respectively.
 In addition, `sim.changed()` is introduced that allows creating triggers from arbitrary signals.
@@ -108,13 +110,21 @@ The following `Simulator` methods have their signatures updated:
 
 The new optional named argument `passive` registers the testbench as passive when true.
 
-Both methods are updated to accept an async function passed as `process`. When the function passed to `process` accepts an argument named `sim`, it will be passed a simulator context.
+Both methods are updated to accept an async function passed as `process`.
+The async function must accept a named argument `sim`, which will be passed a simulator context.
 
 The simulator context have the following methods:
-- `delay(interval=None)`
+- `get(signal)`
+  - Returns the value of `signal` when awaited.
+    When `signal` is a value-castable, the value will be converted through `.from_bits()`. (Pending RFC #51)
+- `set(signal, value)`
+  - Set `signal` to `value` when awaited.
+    When `signal` is a value-castable, the value will be converted through `.const()`.
+- `delay(interval)`
   - Return a trigger object for advancing simulation by `interval` seconds.
-- `tick(domain="sync")`
+- `tick(domain="sync", *, context=None)`
   - Return a trigger object for advancing simulation by one tick of `domain`.
+    When an elaboratable is passed to `context`, `domain` will be resolved from its perspective.
 - `changed(signal, value=None)`
   - Return a trigger object for advancing simulation until `signal` is changed to `value`. `None` is a wildcard and will trigger on any change.
 - `active()`
@@ -124,24 +134,27 @@ The simulator context have the following methods:
 
 A trigger object has the following methods:
 - `until(condition)`
-  - Repeat the trigger until `condition` is true. If `condition` is initially true, `await` will return immediately without advancing simulation.
+  - Repeat the trigger until `condition` is true.
+    `condition` is an arbitrary Amaranth expression.
+    If `condition` is initially true, `await` will return immediately without advancing simulation.
 
-`Value`, `data.View` and `enum.EnumView` have `.get()` and `.set()` methods added.
+~~`Value`, `data.View` and `enum.EnumView` have `.get()` and `.set()` methods added.~~
 
 `Tick()`, `Delay()`, `Active()` and `Passive()` as well as the ability to pass generator coroutines as `process` are deprecated and removed in a future version.
 
 ## Drawbacks
 [drawbacks]: #drawbacks
 
-Reserves two new names on `Value` and value castables. Increase in API surface area and complexity. Churn.
+-  ~~Reserves two new names on `Value` and value castables~~
+- Increase in API surface area and complexity.
+- Churn.
 
 ## Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
 - Do nothing. Keep the existing interface, add `Changed()` alongside `Delay()` and `Tick()`, use `yield from` when calling functions.
 
-- Don't introduce `.get()` and `.set()`. Instead require a value castable and the return value of its `.eq()` to be awaitable so `await value` and `await value.eq(foo)` is possible.
-
+- ~~Don't introduce `.get()` and `.set()`. Instead require a value castable and the return value of its `.eq()` to be awaitable so `await value` and `await value.eq(foo)` is possible.~~
 
 ## Prior art
 [prior-art]: #prior-art
@@ -156,9 +169,11 @@ Other python libraries like [cocotb](https://docs.cocotb.org/en/stable/coroutine
   Simulating sync logic with async reset could be another.
   What would be a good syntax to combine triggers?
 - Is there any other functionality that's natural to have on the simulator context?
+  - (@wanda-phi) `sim.memory_read(memory, address)`, `sim.memory_write(memory, address, value[, mask])`?
 - Is there any other functionality that's natural to have on the trigger object?
   - Maybe a way to skip a given number of triggers? We still lack a way to say «advance by n cycles».
 - Bikeshed all the names.
+  - (@whitequark) We should consider different naming for `active`/`passive`.
 
 ## Future possibilities
 [future-possibilities]: #future-possibilities
--- a/text/0036-async-testbench-functions.md
+++ b/text/0036-async-testbench-functions.md
@@ -18,8 +18,6 @@ A more expressive way to specify trigger/wait conditions allows the condition ch
 
 Passing a simulator context to the testbench function provides a convenient place to gather all simulator operations.
 
-~~Having `.get()` and `.set()` methods provides a convenient way for value castables to implement these in a type-specific manner.~~
-
 ## Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
@@ -53,6 +51,10 @@ class StreamInterface(PureInterface):
 
 `sim.tick()` replaces the existing `Tick()`. It returns a trigger object that either can be awaited directly, or made conditional through `.until()`.
 
+> **Note**
+> This simplified example does not include any way of specifying the clock domain of the interface and as such is only directly applicable to single domain simulations.
+> A way to attach clock domain information to interfaces is desireable, but out of scope for this RFC.
+
 Using this stream interface, let's consider a colorspace converter accepting a stream of RGB values and outputting a stream of YUV values:
 
 ```python
@@ -82,11 +84,12 @@ async def testbench(sim):
 Since `.send()` and `.recv()` invokes `sim.get()` and `sim.set()` that in turn will invoke the appropriate value conversions for a value castable (here `data.View`), it is general enough to work for streams with arbitrary shapes.
 
 `Tick()` and `Delay()` are replaced by `sim.tick()` and `sim.delay()` respectively.
-In addition, `sim.changed()` is introduced that allows creating triggers from arbitrary signals.
+In addition, `sim.changed()` and `sim.edge()` is introduced that allows creating triggers from arbitrary signals.
 These all return a trigger object that can be made conditional through `.until()`.
 
-`Active()` and `Passive()` are replaced by an `passive=False` keyword argument to `.add_process()` and `.add_testbench()`.
-To mark a passive testbench temporarily active, `sim.active()` is introduced, which is used as a context manager:
+`Active()` and `Passive()` are replaced by an `background=False` keyword argument to `.add_testbench()`.
+Processes created through `.add_process()` are always created as background processes.
+To allow a background process to ensure an operation is finished before end of simulation, `sim.critical()` is introduced, which is used as a context manager:
 
 ```python
 async def packet_reader(sim, stream):
@@ -94,10 +97,46 @@ async def packet_reader(sim, stream):
         # Wait until stream has valid data.
         await sim.tick().until(stream.valid)
 
-        # Go active to ensure simulation doesn't end in the middle of a packet.
-        async with sim.active():
+        # Ensure simulation doesn't end in the middle of a packet.
+        async with sim.critical():
             packet = await stream.read_packet()
             print('Received packet:', packet.hex(' '))
+```
+
+When a trigger object is awaited, it'll return the value(s) of the trigger(s), and it can also be used as an async generator to repeatedly await the same trigger.
+Multiple triggers can be combined.
+Consider the following examples:
+
+Combinational adder as a process:
+```python
+a = Signal(); b = Signal(); o = Signal()
+async def adder(sim):
+    async for a_val, b_val in sim.changed(a, b):
+        await sim.set(o, a_val + b_val)
+sim.add_process(adder)
+```
+
+DDR IO buffer as a process:
+```python
+o = Signal(2); pin = Signal()
+async def ddr_buffer(sim):
+    while True: # could be extended to pre-capture next `o` on posedge
+        await sim.negedge()
+        await sim.set(pin, o[0])
+        await sim.posedge()
+        await sim.set(pin, o[1])
+sim.add_process(ddr_buffer)
+```
+
+Flop with configurable edge reset and posedge clock as a process:
+```python
+clk = Signal(); rst = Signal(); d = Signal(); q = Signal()
+def dff(rst_edge):
+    async def process(sim):
+        async for clk_val, rst_val in sim.posedge(clk).edge(rst, rst_edge):
+            await sim.set(q, 0 if rst_val == rst_edge else await sim.get(d))
+    return process
+sim.add_process(dff(rst_edge=0))
 ```
 
 ## Reference-level explanation
@@ -105,47 +144,71 @@ async def packet_reader(sim, stream):
 
 The following `Simulator` methods have their signatures updated:
 
-* `add_process(process, *, passive=False)`
-* `add_testbench(process, *, passive=False)`
+* `add_process(process)`
+* `add_testbench(process, *, background=False)`
 
-The new optional named argument `passive` registers the testbench as passive when true.
+The new optional named argument `background` registers the testbench as a background process when true.
 
 Both methods are updated to accept an async function passed as `process`.
-The async function must accept a named argument `sim`, which will be passed a simulator context.
+The async function must accept an argument `sim`, which will be passed a simulator context.
+(Argument name is just convention, will be passed positionally.)
 
 The simulator context have the following methods:
-- `get(signal)`
-  - Returns the value of `signal` when awaited.
-    When `signal` is a value-castable, the value will be converted through `.from_bits()`. (Pending RFC #51)
-- `set(signal, value)`
-  - Set `signal` to `value` when awaited.
-    When `signal` is a value-castable, the value will be converted through `.const()`.
+- `get(expr: Value) -> int`
+- `get(expr: ValueCastable) -> any`
+  - Returns the value of `expr` when awaited.
+    When `expr` is a value-castable, the value will be converted through `.from_bits()`.
+- `set(expr: Value, value: ConstLike)`
+- `set(expr: ValueCastable, value: any)`
+  - Set `expr` to `value` when awaited.
+    When `expr` is a value-castable, the value will be converted through `.const()`.
+- `memory_read(instance: MemoryInstance, address)`
+  - Read the value from `address` in `instance` when awaited.
+- `memory_write(instance: MemoryInstance, address, value, mask=None)`
+  - Write `value` to `address` in `instance` when awaited. If `mask` is given, only the corresponding bits are written.
 - `delay(interval)`
-  - Return a trigger object for advancing simulation by `interval` seconds.
+  - Create a trigger object for advancing simulation by `interval` seconds.
 - `tick(domain="sync", *, context=None)`
-  - Return a trigger object for advancing simulation by one tick of `domain`.
+  - Create a trigger object for advancing simulation by one tick of `domain`.
     When an elaboratable is passed to `context`, `domain` will be resolved from its perspective.
-- `changed(signal, value=None)`
-  - Return a trigger object for advancing simulation until `signal` is changed to `value`. `None` is a wildcard and will trigger on any change.
-- `active()`
-  - Return a context manager that temporarily marks the testbench as active for the duration.
-- `time()`
-  - Return the current simulation time.
+  - If `domain` is asynchronously reset while this is being awaited, `AsyncReset` is raised.
+- `changed(*signals)`
+  - Create a trigger object for advancing simulation until any signal in `signals` changes.
+- `edge(signal, value)`
+  - Create a trigger object for advancing simulation until `signal` is changed to `value`.
+    `signal` must be a 1-bit signal or a 1-bit slice of a signal.
+- `posedge(signal)`
+- `negedge(signal)`
+  - Aliases for `edge(signal, 1)` and `edge(signal, 0)` respectively.
+  
+  `signal` is changed to `value`. `None` is a wildcard and will trigger on any change.
+- `critical()`
+  - Return a context manager that ensures simulation won't terminate in the middle of the enclosed scope.
 
 A trigger object has the following methods:
+- `__await__()`
+  - Advance simulation and return the value(s) of the trigger(s).
+    - `delay` and `tick` triggers return `True` when they are hit, otherwise `False`.
+    - `changed` and `edge` triggers return the current value of the signals they are monitoring.
+- `__aiter__()`
+  - Return an async generator that repeatedly invokes `__await__()` and yields the returned values.
+- `delay(interval)`
+- `tick(domain="sync", *, context=None)`
+- `changed(*signals)`
+- `edge(signal, value)`
+- `posedge(signal)`
+- `negedge(signal)`
+  - Create a new trigger object by copying the current object and appending another trigger.
 - `until(condition)`
   - Repeat the trigger until `condition` is true.
     `condition` is an arbitrary Amaranth expression.
     If `condition` is initially true, `await` will return immediately without advancing simulation.
-
-~~`Value`, `data.View` and `enum.EnumView` have `.get()` and `.set()` methods added.~~
 
 `Tick()`, `Delay()`, `Active()` and `Passive()` as well as the ability to pass generator coroutines as `process` are deprecated and removed in a future version.
 
 ## Drawbacks
 [drawbacks]: #drawbacks
 
--  ~~Reserves two new names on `Value` and value castables~~
 - Increase in API surface area and complexity.
 - Churn.
 
@@ -153,8 +216,6 @@ A trigger object has the following methods:
 [rationale-and-alternatives]: #rationale-and-alternatives
 
 - Do nothing. Keep the existing interface, add `Changed()` alongside `Delay()` and `Tick()`, use `yield from` when calling functions.
-
-- ~~Don't introduce `.get()` and `.set()`. Instead require a value castable and the return value of its `.eq()` to be awaitable so `await value` and `await value.eq(foo)` is possible.~~
 
 ## Prior art
 [prior-art]: #prior-art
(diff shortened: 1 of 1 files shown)


zyp:
Updated to match the changes in #27.

text/0036-async-testbench-functions.md
```diff
@@ -29,26 +29,28 @@ On the interface class, we can then implement `.send()` and `.recv()` methods li
...
-        await self.valid.set(1)
+        await sim.set(self.valid, 1)
         await sim.tick().until(self.ready)
 
         await sim.tick()
-        await self.valid.set(0)
+        await sim.set(self.valid, 0)
 ```
 
+`sim.get()` and `sim.set()` replaces the existing operations `yield signal` and `yield signal.eq()` respectively.
```
whitequark, line 52:
```suggestion
`await sim.get()` and `await sim.set()` replaces the existing operations `yield signal` and `yield signal.eq()` respectively.
```

whitequark:
I think this should add a function `run_async` that causes any `await`s to be forwarded to an external event loop, like from `asyncio`.

zyp:
I think that is a good idea, but orthogonal enough to this RFC that it deserves its own RFC. I can write it if you'd like.

whitequark:
Sounds good to me!

whitequark:
> `time()`
> Return the current simulation time.

What is the type of the returned value?

zyp:
Presumably seconds as a `float`, for consistency, since that's what's used in e.g. `add_clock()`'s `period` argument. I'm open for other suggestions though.

whitequark:
> Presumably seconds as a `float`, for consistency

Using `float` in `add_clock` was a bad decision and we should probably not multiply it if we can avoid it...

zyp:
In that case I propose we change all simulation time units to seconds as a `Fraction`.

whitequark:
Sounds good to me. It should internally use femtoseconds though.

zyp:
Shouldn't what's used internally be an implementation detail left to the simulation engine? That said, using femtoseconds would throw away the benefit of using a `Fraction` since it's potentially lossy. The period of e.g. `30 MHz` is not an integer femtoseconds.

Is it worth using a more complicated type than `float` if it might be lossy anyway? (FWIW, a Python `float` gives femtosecond resolution up to nine seconds of simulation time or so.)

whitequark:
CXXRTL uses femtoseconds internally (with a custom data type wrapping them into a integral seconds plus fractional second as femtoseconds) and I'd like the rounding error to match between Python and C++.

whitequark:
Some comments in advance of the Monday meeting.

> As an example, let's consider a simple stream interface with valid, ready and data members.

This example has the glaring issue of not having any way of specifying the clock domain at all. This needs to be called out in the RFC text as an avenue for future work; i.e. that the example isn't the outcome directly enabled by the RFC but a desirable future result.

A better solution perhaps is to reframe these functions as not methods on the streams themselves but as helpers that could be eventually put onto the streams. Then you can just have the clock domain as an argument (or ignore it like I did in some single-clock code I worked on recently) and the example motivates the RFC well.

> `add_process(process, *, passive=False)`

Should `add_process` not have `passive=True` by default? In our model, processes are going to be reserved exclusively for behavioral replacement of logic, and simulated logic is functionally passive.

Alternatives:
1. no default `passive=` argument for `add_process` and value must have specified explicitly;
2. `add_process()` is only capable of creating passive-by-default processes.

I particularly like the 2nd alternative since it further cements the distinction in the application of the two APIs, which I would like to make as clear as possible.

> `add_testbench(process, *, passive=False)`

It's not completely clear to me, but when do we want passive testbenches?

If we remove `passive` from both `add_process` and `add_testbench` we end up with a much simpler API conceptually:
- `add_process` replaces some logic behaviorally. In most cases it can be shut down if no testbench is running, however in recognition that e.g. an AXI transaction monitor acts as-if it was logic attached to the bus, but needs to ensure that any transaction proceeds to completion, it can temporarily prevent the simulation from being shut down.

> The async function must accept a named argument sim, which will be passed a simulator context.

Now that we've amended the RFC to have a different API for legacy and async processes we can pass the argument positionally.

> `get(signal)`

This does not just work on signals, right? I think the actual signature is best described by the two overloads:
- `get(expr: Value) -> int`
- `get(expr: ValueCastable) -> any`

> Returns the value of signal when awaited. When signal is a value-castable, the value will be converted through .from_bits(). (Pending RFC #51)

RFC #51 has been accepted.

> `set(signal, value)`

Similarly I would propose:
- `set(expr: Value, value: ConstLike)`
- `set(expr: ValueCastable, value: any)`

(Emphasis on argument names here.)

> `delay(interval)`

I think this needs a better type specification. I don't recall what we currently support so I'll leave it as-is.

Should `delay()` be allowed in `add_process()`? (I think yes; otherwise you could not e.g. simulate a phase shifter, which seems desirable.)

> `tick(domain="sync", *, context=None)`

There are two significant design issues with `tick` (and `Tick` on which it is based). Consider what happens when you call `await tick("domain")` and the domain has its async reset triggered? As it stands, the `await` call returns.

The design issues are:
- The name `tick` (or the `clocked` replacement I proposed earlier) evokes the idea that it triggers on a clock edge, but it triggers on reset edge too.
- There is no straightforward way to distinguish the two cases between each other.

Some strawman proposals I can come up with:
- Forbid `tick()` from triggering on async reset and if this ever happens in a simulation, crash all the waiting processes. (Very rude.)
- Add a parameter to `tick()` declaring whether an async reset edge is expected, crashing the process as above if it's not specified or `False`.
- Make the `await sim.tick()` call return `True` if the domain was async-reset, `False` otherwise.
  - The same but any domain reset counts.
  - This can be ignored too easily.
- Make the call return a enum or something so that you have to use `if await sim.tick() == BikeshedEnum.Clock:` (the precedence does work out correctly).
  - Make the `BikeshedEnum` a `MustUse` class.
  - Only do this by opt-in.
  - Use two functions, one async-reset-aware and one async-reset-unaware.

> `changed(signal, value=None)`

We probably need some way to trigger on up to two signals with specific values (for sync processes), or an arbitrary number of values without specific values (for comb processes).

Strawman proposals:
- `changed((signal1, signal2, ...), value=(value1, value2, ...))`; a straightforward extension of the current API.
  - Ugly.
- `changed(signal, value=None).or_changed(signal, value=None)`; chaining like with `until`
  - Really verbose and ugly for comb processes.
  - Creates what is essentially a simulation-specific sub-language with entirely new syntax and unclear semantics.
- `changed(signal)` or  `changed(signal == value)` or `changed((signal1 == value1) | (signal2 == value2))`
  - Worse syntax than Verilog.
  - Doesn't even match semantics of what's happening.
- `changed(signal, ...)` or `posedge(signal, ...)` or `negedge(signal, ...)`, chainable
  - Composable, clearly defined semantics that doesn't borrow syntax from the core language in a misleading way.
  - Obnoxious for processes generic over edge type, without an obvious way to solve this as long as multiple signals can be provided. (`edge((signal, 1), ...)` has all the problems of the first option.)
- `changed(signal, ...)` (without allowing a value) + `tick()` (without any way to trigger on an edge that doesn't involve a domain)
  - Seems limiting, as e.g. a DDR register implementation may need to trigger on a clock both ways.

I think my final proposal is a variation on some of the above, namely the combination of:
- `async for values in sim.changed(signals: Iterable[Signal])`: perfect for implementing comb process replacements, allows the simulator to amortize the cost of waiting on `signals` by registering event listeners and leaving them there.
- `await sim.edge(clk, 1).edge(rst, 0)`: perfect for implementing sync processes or variations on the theme, like DDR buffers.
  - Could have `sim.posedge(clk)` as an alias.
  - First argument must be a 1-bit signal or a 1-bit slice of a signal.
  - Returns the values of all registered triggers.

Examples:

1. Combinational adder as a process:

   ```py
   a = Signal(); b = Signal(); o = Signal()
   async def adder(sim):
       async for a_val, b_val in sim.changed(a, b):
           await sim.set(o, a_val + b_val)
   sim.add_process(adder)
   ```

2. DDR IO buffer as a process:

   ```py
   o = Signal(2); pin = Signal()
   async def ddr_buffer(sim):
       while True: # could be extended to pre-capture next `o` on posedge
           await sim.negedge()
           await sim.set(pin, o[0])
           await sim.posedge()
           await sim.set(pin, o[1])
   sim.add_process(ddr_buffer)
   ```

3. Flop with configurable edge reset and posedge clock as a process:

   ```py
   clk = Signal(); rst = Signal(); d = Signal(); q = Signal()
   def dff(rst_edge):
       async def process(sim):
           while True:
               clk_val, rst_val = await sim.posedge(clk).edge(rst, rst_edge)
               await sim.set(q, 0 if rst_val == rst_edge else await sim.get(d))
       return process
   sim.add_process(dff(rst_edge=0))
   ```

> `active()`

Proposed new name: `critical()`. This is a name that will catch the eye of the reader, but that's as it should: it's not normal at all for `add_process` to block the simulation from exiting. Downside: "critical section" isn't widely used in Python parlance.

Alternative: `lock()`. Downside: too generic.

> `time()`

Given the lack of consensus on what it should return in the current codebase (with the proposed solution being a new type, `amaranth.sim.Time` or something like that), I propose descoping this feature from this RFC. There's plenty to go over already, and I think it's quite critical that we ship this soon.

> It should be possible to combine triggers

Addressed in this comment.

> (@wanda-phi) sim.memory_read(memory, address), sim.memory_write(memory, address, value[, mask])?

Seems OK to me to include with this name. People who use autocompletion on `sim.` may get confused but they'll also get confused by `sim.memory_instance_read`, and adding a type annotation on the first argument (which should be `instance: MemoryInstance`) should do much to address this confusion.

`sim` deals with core types, not library types, after all.

> Maybe a way to skip a given number of triggers? We still lack a way to say «advance by n cycles».

This is trivial to work around with the current syntax:

```py
for _ in range(5): await sim.tick()
```

Adding syntax for this would be an optimization, but pysim doesn't need it and cxxsim isn't anywhere near the stage where it can be used. I think this concern can be safely ignored in this RFC (and moved to "Future work").

zyp:
> It's not completely clear to me, but when do we want passive testbenches?

I've got some stuff where I've got one testbench that pushes packets into one end of a pipeline and another that reads out and prints packets on the other end, and it's convenient for the latter to be passive so I don't have to tell it when there's no more packets coming.

> > `delay(interval)`
>
> I think this needs a better type specification. I don't recall what we currently support so I'll leave it as-is.

IIRC it currently takes seconds as a float. I figure that's good enough until we've got `amaranth.sim.Time`.

> Should `delay()` be allowed in `add_process()`? (I think yes; otherwise you could not e.g. simulate a phase shifter, which seems desirable.)

Yes, I see no reason why not. I was also thinking we could disallow some triggers depending on the process type, but I haven't come up with any convincing arguments for doing so.

> I think my final proposal is a variation on some of the above…

I took this and generalized it so that we've still got a common trigger object and so that they can all be combined.

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+In addition, `sim.changed()` and `sim.edge()` is introduced that allows creating triggers from arbitrary signals.
+These all return a trigger object that can be made conditional through `.until()`.
+
+`Active()` and `Passive()` are replaced by an `background=False` keyword argument to `.add_testbench()`.
+Processes created through `.add_process()` are always created as background processes.
+To allow a background process to ensure an operation is finished before end of simulation, `sim.critical()` is introduced, which is used as a context manager:
+
+```python
+async def packet_reader(sim, stream):
+    while True
```
whitequark, line 96:
```suggestion
    while True:
```

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+The simulator context have the following methods:
+- `get(expr: Value) -> int`
+- `get(expr: ValueCastable) -> any`
+  - Returns the value of `expr` when awaited.
+    When `expr` is a value-castable, the value will be converted through `.from_bits()`.
+- `set(expr: Value, value: ConstLike)`
+- `set(expr: ValueCastable, value: any)`
+  - Set `expr` to `value` when awaited.
+    When `expr` is a value-castable, the value will be converted through `.const()`.
+- `memory_read(instance: MemoryInstance, address)`
```
whitequark, line 165:
```suggestion
- `memory_read(instance: MemoryInstance, address: int)`
```

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+  - Read the value from `address` in `instance` when awaited.
+- `memory_write(instance: MemoryInstance, address, value, mask=None)`
```
whitequark, line 167:
```suggestion
- `memory_write(instance: MemoryInstance, address: int, value: int, mask:int = None)`
```

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+  - Write `value` to `address` in `instance` when awaited. If `mask` is given, only the corresponding bits are written.
+- `delay(interval)`
```
whitequark, line 169:
```suggestion
- `delay(interval: float)`
```

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+  - Create a trigger object for advancing simulation by one tick of `domain`.
+    When an elaboratable is passed to `context`, `domain` will be resolved from its perspective.
+  - If `domain` is asynchronously reset while this is being awaited, `AsyncReset` is raised.
+- `changed(*signals)`
+  - Create a trigger object for advancing simulation until any signal in `signals` changes.
+- `edge(signal, value)`
+  - Create a trigger object for advancing simulation until `signal` is changed to `value`.
+    `signal` must be a 1-bit signal or a 1-bit slice of a signal.
+- `posedge(signal)`
+- `negedge(signal)`
```
whitequark, line 181: Bikeshed: should we go for `posedge` (Verilog convention) or `pos_edge` (Python convention)?

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+  - Aliases for `edge(signal, 1)` and `edge(signal, 0)` respectively.
+  
+  `signal` is changed to `value`. `None` is a wildcard and will trigger on any change.
```
whitequark, line 184: I think this line got lost

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+- `critical()`
+  - Return a context manager that ensures simulation won't terminate in the middle of the enclosed scope.
```
whitequark, line 186: That ensures simulation won't terminate *because run() ran out of non-background processes*. I think this may be better put as "if the current process is a background process, `async with critical()` makes it a non-background process for the duration of the statement".

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+
+A trigger object has the following methods:
+- `__await__()`
+  - Advance simulation and return the value(s) of the trigger(s).
+    - `delay` and `tick` triggers return `True` when they are hit, otherwise `False`.
+    - `changed` and `edge` triggers return the current value of the signals they are monitoring.
+- `__aiter__()`
+  - Return an async generator that repeatedly invokes `__await__()` and yields the returned values.
```
whitequark, line 194: I think this should be described in terms of simulation behavior, not implementation (since the implementation may want to do something more efficient!)

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+- `delay(interval)`
+- `tick(domain="sync", *, context=None)`
+- `changed(*signals)`
+- `edge(signal, value)`
+- `posedge(signal)`
+- `negedge(signal)`
```
whitequark, line 200: If I construct something like `await sim.delay(interval).posedge(signal)`, what is the behavior?

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+  - Create a new trigger object by copying the current object and appending another trigger.
+- `until(condition)`
```
whitequark, line 202: What does this function return? Is it another trigger object? What does `await sim.delay(1).until(x).delay(1)` mean?

whitequark:
Observation:

```py
    async for a_val, b_val in sim.changed(a, b):
    async for clk_val, rst_val in sim.posedge(clk).posedge(rst):
```

is literally just

```v
    always @(a, b)
    always @(posedge clk, posedge rst)
```

We have reinvented Verilog in our simulation sub-language. How horrible :D

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,235 @@
...
+- `memory_read(instance: MemoryInstance, address)`
+  - Read the value from `address` in `instance` when awaited.
+- `memory_write(instance: MemoryInstance, address, value, mask=None)`
+  - Write `value` to `address` in `instance` when awaited. If `mask` is given, only the corresponding bits are written.
+- `delay(interval)`
+  - Create a trigger object for advancing simulation by `interval` seconds.
+- `tick(domain="sync", *, context=None)`
+  - Create a trigger object for advancing simulation by one tick of `domain`.
+    When an elaboratable is passed to `context`, `domain` will be resolved from its perspective.
+  - If `domain` is asynchronously reset while this is being awaited, `AsyncReset` is raised.
```
whitequark, line 174:
```suggestion
  - If `domain` is asynchronously reset while this is being awaited, `amaranth.sim.AsyncReset` is raised.
```

text/0036-async-testbench-functions.md
zyp: If you combine a `delay` with another trigger, you get a timeout-like behavior. Means you could do something like:
```python
_, timeout = await sim.posedge(signal).delay(0.01)
assert not timeout, "signal must go positive within 10ms"
```

I thought about this after I pushed the previous change, and that `edge` should probably also return whether it was hit like `tick` and `delay`, and not the actual value like `changed`, since we already know the only value it could be.

Combining two delays naturally means that the shorter one will always trigger first.
If they've got the same interval, it's an undefined race like when two clocks tick at the same time step.
While the semantics are clear enough, I don't see a good use for this, so we could forbid adding a second delay trigger to a trigger object that already has one.

whitequark: I can't see implementing this being easy.

zyp: At a glance, `BaseSimulation` needs another method to cancel a delay trigger added by `wait_interval()`, pysim (and of course cxxsim) needs to implement it and `PyCoroProcess` needs to call it when it's doing `clear_triggers()`.

whitequark: Actually, I guess you could just track the closest `.delay` trigger there is inside the trigger object since it will always fire before the other ones.

whitequark:
> > Should `delay()` be allowed in `add_process()`? (I think yes; otherwise you could not e.g. simulate a phase shifter, which seems desirable.)
> 
> Yes, I see no reason why not. I was also thinking we could disallow some triggers depending on the process type, but I haven't come up with any convincing arguments for doing so.

Interestingly, this means that `Settle` is not *truly* removed, as a zero time delay does basically the same thing. (With the advantage of not implying that signals will be truly settled and there won't be any race conditions...)

text/0036-async-testbench-functions.md
```diff
@@ -162,16 +162,16 @@ The simulator context have the following methods:
...
+- `memory_read(instance: MemoryInstance, address: int)`
   - Read the value from `address` in `instance` when awaited.
-- `memory_write(instance: MemoryInstance, address, value, mask=None)`
+- `memory_write(instance: MemoryInstance, address: int, value: int, mask:int = None)`
   - Write `value` to `address` in `instance` when awaited. If `mask` is given, only the corresponding bits are written.
-- `delay(interval)`
+- `delay(interval: float)`
   - Create a trigger object for advancing simulation by `interval` seconds.
 - `tick(domain="sync", *, context=None)`
   - Create a trigger object for advancing simulation by one tick of `domain`.
```
whitequark, line 172: I think this should be further clarified to:

> - Create a trigger object for advancing simulation until the next active edge of the `domain` clock.

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+    - `changed` triggers return the current value of the signals they are monitoring.
+- `__aiter__()`
+  - Return an async generator that is equivalent to repeatedly awaiting the trigger object in an infinite loop.
+- `delay(interval: float)`
+- `tick(domain="sync", *, context=None)`
+- `changed(*signals)`
+- `edge(signal, value)`
+- `posedge(signal)`
+- `negedge(signal)`
+  - Create a new trigger object by copying the current object and appending another trigger.
```
whitequark, line 200:
```suggestion
  - Create a new trigger object by copying the current object and appending another trigger.
  - Awaiting the returned trigger object pauses the process until the first of the combined triggers hit, i.e. the triggers are combined using OR semantics.
```

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+- `until(condition)`
+  - Repeat the trigger until `condition` is true.
+    `condition` is an arbitrary Amaranth expression.
+    If `condition` is initially true, `await` will return immediately without advancing simulation.
+    The return value is an unspecified awaitable with `await` as the only defined operation.
```
whitequark, line 205: The return value of the awaitable is what?

What happens if I did something like...

```
await sim.posedge(clk).posedge(rst).until(2)
```

Do I learn exactly what was the first trigger, or is that information simply lost?

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+  - Aliases for `edge(signal, 1)` and `edge(signal, 0)` respectively.
+- `critical()`
+  - Context manager.
+    If the current process is a background process, `async with sim.critical():` makes it a non-background process for the duration of the statement.
+
+A trigger object has the following methods:
+- `__await__()`
+  - Advance simulation and return the value(s) of the trigger(s).
+    - `delay`, `tick` and `edge` triggers return `True` when they are hit, otherwise `False`.
+    - `changed` triggers return the current value of the signals they are monitoring.
```
whitequark, line 191: What happens if multiple triggers are hit simultaneously? E.g.

```py
x = await sim.posedge(clk).posedge(clk)
```

Does this return `(True, True)`, or `(True, False)`, or `(False, True)`, or unspecified / any of the above?

What about this snippet, assuming f_clk = 1e-6?

```py
await sim.posedge(clk)
x = await sim.posedge(clk).delay(1e-6)
```
Does this guarantee `(True, True)` still? I think this might be quite difficult to achieve reliably with every possible simulation engine.

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+As an example, let's consider a simple stream interface with `valid`, `ready` and `data` members.
+On the interface class, we can then implement `.send()` and `.recv()` methods like this:
+
+```python
+class StreamInterface(PureInterface):
+    async def recv(self, sim):
```
whitequark, line 29: To be honeat I feel that as written this example is still fairly misleading as not only to capabilities of this RFC but also currently in-pipeline RFCs and it would be better to rewrite these methods as function (non-method) simulation helpers, to avoid any ambiguity.

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The following `Simulator` methods have their signatures updated:
+
+* `add_process(process)`
+* `add_testbench(process, *, background=False)`
+
+The new optional named argument `background` registers the testbench as a background process when true.
```
whitequark, line 150: This doesn't explain when processes created by `add_process()` become background or not. They become background iff an async generator function is passed as `process`, which is quite not obvious!

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+  - Create a trigger object for advancing simulation by `interval` seconds.
+- `tick(domain="sync", *, context=None)`
+  - Create a trigger object for advancing simulation until the next active edge of the `domain` clock.
+    When an elaboratable is passed to `context`, `domain` will be resolved from its perspective.
+  - If `domain` is asynchronously reset while this is being awaited, `amaranth.sim.AsyncReset` is raised.
+- `changed(*signals)`
+  - Create a trigger object for advancing simulation until any signal in `signals` changes.
+- `edge(signal, value)`
+  - Create a trigger object for advancing simulation until `signal` is changed to `value`.
+    `signal` must be a 1-bit signal or a 1-bit slice of a signal.
```
whitequark, line 179: What should `value` be?

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+
+A trigger object has the following methods:
+- `__await__()`
+  - Advance simulation and return the value(s) of the trigger(s).
+    - `delay`, `tick` and `edge` triggers return `True` when they are hit, otherwise `False`.
+    - `changed` triggers return the current value of the signals they are monitoring.
+- `__aiter__()`
+  - Return an async generator that is equivalent to repeatedly awaiting the trigger object in an infinite loop.
+- `delay(interval: float)`
+- `tick(domain="sync", *, context=None)`
```
whitequark, line 195: I think `tick()` should not be a combinator, chiefly because if you have something like...

```
await sim.tick("doma").tick("domb")
```

and an `AsyncReset` exception is raised, you have no way of figuring out where the exception came from.

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+
+Both methods are updated to accept an async function passed as `process`.
+The async function must accept an argument `sim`, which will be passed a simulator context.
+(Argument name is just convention, will be passed positionally.)
+
+The simulator context have the following methods:
+- `get(expr: Value) -> int`
+- `get(expr: ValueCastable) -> any`
+  - Returns the value of `expr` when awaited.
+    When `expr` is a value-castable, the value will be converted through `.from_bits()`.
```
wanda-phi, line 160:
```suggestion
    When `expr` is a value-castable, and its `shape()` is a `ShapeCastable`, the value will be converted through the shape's `.from_bits()`. Otherwise, a plain integer is returned.
```

this is necessary, as you can have `ValueCastable`s whose shape is just a plain `Shape`

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+- `set(expr: Value, value: ConstLike)`
+- `set(expr: ValueCastable, value: any)`
+  - Set `expr` to `value` when awaited.
+    When `expr` is a value-castable, the value will be converted through `.const()`.
```
wanda-phi, line 164:
```suggestion
    When `expr` is a value-castable, and its `shape()` is a `ShapeCastable`, the value will be converted through the shape's `.const()`. Otherwise, it must be a const-castable `ValueLike`.
```

ditto

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+- `memory_read(instance: MemoryInstance, address: int)`
```
wanda-phi, line 165:
```suggestion
- `memory_read(instance: MemoryIdentity, address: int)`
```

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+  - Read the value from `address` in `instance` when awaited.
+- `memory_write(instance: MemoryInstance, address: int, value: int, mask:int = None)`
```
wanda-phi, line 167:
```suggestion
- `memory_write(instance: MemoryIdentity, address: int, value: int, mask:int = None)`
```

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,236 @@
...
+  - Write `value` to `address` in `instance` when awaited. If `mask` is given, only the corresponding bits are written.
```
wanda-phi, line 168:
```suggestion
  - Write `value` to `address` in `instance` when awaited. If `mask` is given, only the corresponding bits are written.
    Like `MemoryInstance`, these two functions are an internal interface that will be usually only used via `lib.Memory`. It comes without a stability guarantee.
```

wanda-phi:
General comment: a new simulation interface also needs to be defined for `lib.memory.Memory`, wrapping the `sim.memory_*` primitives.

text/0036-async-testbench-functions.md
zyp: We can put the domain as an attribute on `AsyncReset`.

wanda-phi:
After thinking a bit on the `.tick()` combinator interactions, we find the complexity of `AsyncReset` handling concerning. We likewise find the `.until()` interaction with combinators to be somewhat complex.

Since `tick()` is unlikely to be combined with other combinators, we are considering the following design:

- `.delay()`, `.edge()` with its variants, and `.changed()` remain as combinable trigger objects; they cannot [?] be used with `.until()`; they provide a low-level interface to create weird primitives
- `.tick()` cannot be combined with any other trigger, including another `.tick()`; however, it can be combined with `.until()` and `.repeat(num: int)`; it provides a high-level interface to interact with a single clock domain

wanda-phi:
I'd also like to again raise the point that I brought up on the last meeting, and that apparently got lost.

We have effectively two kinds of processes, in which `sim` awaitables look the same, but have different semantics. We intend that users create helper simulation functions on elaboratables and whatnot that call these awaitables. It is overwhelmingly likely that any given function will be designed to work on only one of the two kinds of process, and it is likely that someone will accidentally call such a function from the wrong kind of process, and end up with incomprehensible bugs. How do we prevent such a mistake?

text/0036-async-testbench-functions.md
```diff
@@ -147,62 +149,86 @@ The following `Simulator` methods have their signatures updated:
...
+    - The current process is a testbench process added by `add_testbench`.
+
+A domain trigger object has the following methods:
+- `__await__()`
+  - Advance simulation. No value is returned.
+- `until(condition)`
+  - Repeat the trigger until `condition` is true.
+    `condition` is an arbitrary Amaranth expression.
+    If `condition` is initially true, `await` will return immediately without advancing simulation.
+    The return value is an unspecified awaitable with `await` as the only defined operation.
```
whitequark, line 211: I think this isn't specified enough--for example, is mutable state created within `until`, or is it stateless? Same for `repeat`.

text/0036-async-testbench-functions.md
```diff
@@ -147,62 +149,86 @@ The following `Simulator` methods have their signatures updated:
...
+- `delay(interval: float)`
+  - Create a combinable trigger object for advancing simulation by `interval` seconds.
 - `changed(*signals)`
-  - Create a trigger object for advancing simulation until any signal in `signals` changes.
+  - Create a combinable trigger object for advancing simulation until any signal in `signals` changes.
 - `edge(signal, value)`
-  - Create a trigger object for advancing simulation until `signal` is changed to `value`.
+  - Create a combinable trigger object for advancing simulation until `signal` is changed to `value`.
     `signal` must be a 1-bit signal or a 1-bit slice of a signal.
+    `value` is a boolean where true indicates rising edge and false indicates falling edge.
```
whitequark, line 190: Is an `int` also acceptable? The code below suggests that yes, this statement is conflicting with it. (So I guess the type annotation would be `value: int` and also `value in (0, 1)` is required.)

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,262 @@
...
+    `signal` must be a 1-bit signal or a 1-bit slice of a signal.
+    `value` is a boolean where true indicates rising edge and false indicates falling edge.
+- `posedge(signal)`
+- `negedge(signal)`
+  - Aliases for `edge(signal, 1)` and `edge(signal, 0)` respectively.
+- `critical()`
+  - Context manager.
+    If the current process is a background process, `async with sim.critical():` makes it a non-background process for the duration of the statement.
+
+The `ProcessType` enum has the following members:
```
whitequark, line 198: I don't like the idea that we are giving the ability to *branch* on this rather than *assert* on this. It seems suspiciously powerful for the requested feature.

I think a `@sim.testbench_helper` decorator should work just as well for the purpose of denying its use from `add_process` requested by @wanda-phi:

```
class StreamInterface:
    @sim.testbench_helper
    async def recv(self, sim) -> ConstLike:
        ...
```

zyp: I think attempting to disallow branching is just going to invite even more cursed hacks if somebody really wants to.

```python
async def is_testbench():
    @sim.testbench_helper
    async def foo():
        return True
    try:
        return await foo()
    except:
        return False
```

whitequark: I'm OK with that because then I want to hear from them and learn why they want it.

text/0036-async-testbench-functions.md
whitequark: Also, what does awaiting this unspecified awaitable return?

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,262 @@
...
+- `__await__()`
+  - Advance simulation. No value is returned.
+- `until(condition)`
+  - Repeat the trigger until `condition` is true.
+    `condition` is an arbitrary Amaranth expression.
+    If `condition` is initially true, `await` will return immediately without advancing simulation.
+    The return value is an unspecified awaitable with `await` as the only defined operation.
+- `repeat(times: int)`
+  - Repeat the trigger `times` times.
+    The return value is an unspecified awaitable with `await` as the only defined operation.
```
whitequark, line 214: Is `repeat(0)` allowed? (I.e. is it `times > 0` or `times >= 0`?)

What does awaiting this unspecified awaitable return?

text/0036-async-testbench-functions.md
```diff
@@ -0,0 +1,276 @@
...
+        await sim.set(o, a_val + b_val)
+sim.add_process(adder)
+```
+
+DDR IO buffer as a process:
+```python
+o = Signal(2); pin = Signal()
+async def ddr_buffer(sim):
+    while True: # could be extended to pre-capture next `o` on posedge
+        await sim.negedge()
```
whitequark, line 130: I think this isn't valid code--needs to be `await sim.negedge(clk)` or something.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-18#1710781499-1710782625;) this RFC on the 2024-03-18 weekly meeting. The disposition was to **merge**. Unresolved questions were resolved as follows:
- Use `posedge` and `negedge` names (over `pos_edge` and `neg_edge`).

PR #38, opened 2023-12-11
Component signature immutability

whitequark:


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-12-11#1702315549-1702316369;) this RFC on the 2023-12-11 weekly meeting. The dispositon was to **merge**, with the addition that `Component.__init__` should also accept a dict of members.

PR #40, opened 2023-12-22
Add an RFC for arbitrary `Memory` shapes.

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/arbitrary-memory-shape/text/0040-arbitrary-memory-shape.md)

### DIFF
--- a/text/0040-arbitrary-memory-shape.md
+++ b/text/0040-arbitrary-memory-shape.md
@@ -1,6 +1,6 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Start Date: 2024-01-22
 - RFC PR: [amaranth-lang/rfcs#40](https://github.com/amaranth-lang/rfcs/pull/40)
-- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+- Amaranth Issue: [amaranth-lang/amaranth#1048](https://github.com/amaranth-lang/amaranth/issues/1048)
 
 # Arbitrary `Memory` shapes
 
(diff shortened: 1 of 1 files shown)


text/0040-arbitrary-memory-shape.md
```diff
@@ -0,0 +1,76 @@
...
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- This could also be accomplished by adding a wrapper around `Memory`.
+  - A wrapper would result in more code to maintain than simply updating `Memory`, since both the memory object itself and the port objects would have to be wrapped.
+
+## Prior art
+[prior-art]: #prior-art
+
+None.
```
whitequark, line 66: Arrays in any high level language are prior art here.

zyp: Good point, I weren't thinking general enough.

stafverhaegen-chipflow:
Would it also be made possible to specify granularity of the WritePort with arbitrary shape ?

whitequark:
> Would it also be made possible to specify granularity of the WritePort with arbitrary shape ?

I think this should be possible iff the shape is an `ArrayLayout`.

zyp:
Since we're talking about granularity for write enables and not width conversions, there's not really any need to enforce a uniform element shape, and I'm not sure it's worthwhile to try enforcing that the granularity aligns with natural subdivisions of a shape-castable either. We'd end up with a bunch of special casing for `lib.data` that's not necessarily even correct.

Consider something like a `StructLayout({'a': 1, 'b': 7, 'c': 8, 'd': 16})`. Depending on the application, it could make perfect sense to use this with `granularity=8` even though some fields are packed together and some fields spans a multiple.

whitequark:
I feel that it's important to not blindly allow a case that results in arbitrarily torn writes. I would say that if you want granularity to be less than the full word, it has to be a `Shape` and not a shape-castable then. We can always relax this later, but not the other way around, and I think a conservative choice is wise here.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-01-22#1705943777-1705945425;) this RFC on the 2024-01-22 weekly meeting. The disposition was to **merge** with the following changes
- `Memory.width` is kept and becomes a read-only wrapper for `Shape.cast(self.shape).width`;
- `Memory.depth` becomes read-only.

There was consensus that allowing granularity to be specified for `lib.data.ArrayLayout` shapes is desirable, but as `Memory` currently lives in `hdl` and not `lib` this is not possible at the moment.

@zyp Could you please create a tracking issue for this RFC and insert that and the starting date in the RFC text?

PR #41, opened 2023-12-22
Add an RFC for fixed point types.

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/fixed-point-types/text/0041-fixed-point.md)

### DIFF
--- a/text/0041-fixed-point.md
+++ b/text/0041-fixed-point.md
@@ -25,41 +25,55 @@ TODO
 ## Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-This RFC proposes a library addition `amaranth.lib.fixedpoint` with the following contents:
+This RFC proposes a library addition `amaranth.lib.fixed` with the following contents:
 
-`FixedPoint` is a `ShapeCastable` subclass.
+`fixed.Shape` is a `ShapeCastable` subclass.
 The following operations are defined on it:
 
-- `FixedPoint(f_width, /, *, signed)`: Create a `FixedPoint` with `f_width` fractional bits.
-- `FixedPoint(i_width, f_width, /, *, signed)`: Create a `FixedPoint` with `i_width` integer bits and `f_width` fractional bits.
-- `FixedPoint.cast(shape)`: Cast `shape` to a `FixedPoint` instance.
+- `fixed.Shape(f_width, /, *, signed)`: Create a `fixed.Shape` with zero integer bits and `f_width` fractional bits.
+- `fixed.Shape(i_width, f_width, /, *, signed)`: Create a `fixed.Shape` with `i_width` integer bits and `f_width` fractional bits.
+  - The sign bit is not included in `i_width` or `f_width`, so a `fixed.Shape(7, 8, signed=True)` will be 16 bits wide.
+- `fixed.Shape.cast(shape, f_width=0)`: Cast `shape` to a `fixed.Shape` instance.
 - `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
-- `.const(value)`: Create a fixed point constant from an `int` or `float`, rounded to the closest representable value.
+- `.const(value)`: Create a `fixed.Const` from `value`.
 - `.as_shape()`: Return the underlying `Shape`.
-- `.__call__(target)`: Create a `FixedPointValue` over `target`.
+- `.__call__(target)`: Create a `fixed.Value` over `target`.
 
-`Q(*args)` is an alias for `FixedPoint(*args, signed=True)`.
+`SQ(*args)` is an alias for `fixed.Shape(*args, signed=True)`.
 
-`UQ(*args)` is an alias for `FixedPoint(*args, signed=False)`.
+`UQ(*args)` is an alias for `fixed.Shape(*args, signed=False)`.
 
-`FixedPointValue` is a `ValueCastable` subclass.
+`fixed.Value` is a `ValueCastable` subclass.
 The following operations are defined on it:
 
-- `FixedPointValue(shape, target)`: Create a `FixedPointValue` with `shape` over `target`.
-- `FixedPointValue.cast(value)`: Cast `value` to a `FixedPointValue`.
+- `fixed.Value(shape, target)`: Create a `fixed.Value` with `shape` over `target`.
+- `fixed.Value.cast(value, f_width=0)`: Cast `value` to a `fixed.Value`.
 - `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
-- `.shape()`: Return the `FixedPoint` this was created from.
+- `.shape()`: Return the `fixed.Shape` this was created from.
 - `.as_value()`: Return the underlying value.
 - `.eq(value)`: Assign `value`.
-  - If `value` is a `FixedPointValue`, the precision will be extended or rounded as required.
-  - If `value` is an `int` or `float`, the value will be rounded to the closest representable value.
   - If `value` is a `Value`, it'll be assigned directly to the underlying `Value`.
-- `.round(f_width=0)`: Return a new `FixedPointValue` with precision changed to `f_width`, rounding as required.
+  - If `value` is an `int` or `float`, it'll be cast to a `fixed.Const` first.
+  - If `value` is a `fixed.Value`, the precision will be extended or rounded as required.
+- `.round(f_width=0)`: Return a new `fixed.Value` with precision changed to `f_width`, rounding as required.
 - `.__add__(other)`, `.__radd__(other)`, `.__sub__(other)`, `.__rsub__(other)`, `.__mul__(other)`, `.__rmul__(other)`: Binary arithmetic operations.
-  - If `other` is a `Value` or an `int`, it'll be cast to a `FixedPointValue` first.
+  - If `other` is a `Value`, it'll be cast to a `fixed.Value` first.
+  - If `other` is an `int`, it'll be cast to a `fixed.Const` first.
   - If `other` is a `float`: TBD
-  - The result will be a new `FixedPointValue` with enough precision to hold any resulting value without rounding or overflowing.
+  - The result will be a new `fixed.Value` with enough precision to hold any resulting value without rounding or overflowing.
+- `.__lshift__(other)`, `.__rshift__(other)`: Bit shift operations.
 - `.__neg__()`, `.__pos__()`, `.__abs__()`: Unary arithmetic operations.
+
+`fixed.Const` is a `fixed.Value` subclass.
+The following additional operations are defined on it:
+
+- `fixed.Const(value, shape=None)`: Create a `fixed.Const` from `value`. `shape` must be a `fixed.Shape` if specified.
+  - If `value` is an `int` and `shape` is not specified, the smallest shape that will fit `value` will be selected.
+  - If `value` is a `float` and `shape` is not specified, the smallest shape that gives a perfect representation will be selected.
+    If `shape` is specified, `value` will be rounded to the closest representable value first.
+- `.as_integer_ratio()`: Return the value represented as an integer ratio `tuple`.
+- `.as_float()`: Return the value represented as a `float`.
+- Operators are extended to return a `fixed.Const` if all operands are constant.
 
 ## Drawbacks
 [drawbacks]: #drawbacks
@@ -69,18 +83,23 @@ TBD
 ## Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
-- `FixedPointValue.eq()` could cast a `Value` to a `FixedPointValue` first, and thereby shift the value to the integer part instead of assigning directly to the underlying value.
-  However, `Value.eq()` would always grab the underlying value of a `FixedPointValue`, and for consistency both `.eq()` operations should behave in the same manner.
+- `fixed.Value.eq()` could cast a `Value` to a `fixed.Value` first, and thereby shift the value to the integer part instead of assigning directly to the underlying value.
+  However, `Value.eq()` would always grab the underlying value of a `fixed.Value`, and for consistency both `.eq()` operations should behave in the same manner.
   - If we wanted to go the other way, this RFC could be deferred until another RFC allowing `ValueCastable` to override reflected `.eq()` have been merged.
     However, having to explicitly do `value.eq(fp_value.round())` when rounding is desired is arguably preferable to having `value.eq(fp_value)` do an implicit rounding.
 
-- Unlike `.eq()`, it makes sense for arithmetic operations to cast a `Value` to `FixedPointValue`.
+- Unlike `.eq()`, it makes sense for arithmetic operations to cast a `Value` to `fixed.Value`.
   Multiplying an integer with a fixedpoint constant and rounding the result back to an integer is a reasonable and likely common thing to want to do.
+
+- There's two slightly different [Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) definitions, namely whether the bit counts includes the sign bit or not.
+  Not having the sign bit included seems more common, and has the advantage that a number has the same fractional precision whether `i_width` is 0 or not.
+
+- While Q notation names the signed type `Q`, it's more consistent for Amaranth to use `SQ` since other Amaranth types defaults to unsigned.
 
 ## Prior art
 [prior-art]: #prior-art
 
-[Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) is a common and convenient way to specify floating point types.
+[Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) is a common and convenient way to specify fixed point types.
 
 ## Unresolved questions
 [unresolved-questions]: #unresolved-questions
@@ -91,12 +110,23 @@ TBD
   - We could use the same width for `other` as for `self`, adjusted to the appropriate exponent for the value.
   - We could outright reject it, requiring the user to explicitly specify precision like e.g. `value * Q(15).const(1 / 3)`.
 
-- There's two slightly different [Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) definitions, namely whether the bit counts includes the sign bit or not.
-  `UQ(15)` and `UQ(7, 8)` would be 15 bits in either convention, but `Q(15)` and `Q(7, 8)` would be 15 or 16 bits depending on the convention. Which do we pick?
-
 - Are there any other operations that would be good to have?
 
-- Bikeshed all the names.
+- Are there any operations that would be good to *not* have?
+  - This API (`fixed.Shape.cast()`) seems confusing and difficult to use. Should we expose it at all? (@whitequark)
+
+- `Decimal` and/or `Fraction` support?
+  - This could make sense to have, but both can represent values that's not representable as binary fixed point.
+    On the other hand, a Python `float` can perfectly represent any fixed point value up to a total width of 53 bits and any `float` value is perfectly representable as fixed point.
+
+- Name all the things.
+  - Library name:
+    - Bikeshed: `lib.fixed`, `lib.fixnum`. (@whitequark)
+  - Type names:
+    - `fixed.Shape` and `fixed.Value` are one option, though I can see why others may object to it. (@whitequark)
+  - I feel like the `i_width` and `f_width` names are difficult enough to read that it's of more importance than bikeshedding to come up with something more readable. (@whitequark)
+    - `.int_bits`, `.frac_bits`?
+    - cursed option: `int, frac = x.width`?
 
 ## Future possibilities
 [future-possibilities]: #future-possibilities
--- a/text/0041-fixed-point.md
+++ b/text/0041-fixed-point.md
@@ -37,7 +37,6 @@ The following operations are defined on it:
 - A `fixed.Shape` may be constructed using the following aliases:
     - `SQ(i_bits, f_bits)` is an alias for `fixed.Shape(signed(i_bits + f_bits), f_bits)`.
     - `UQ(i_bits, f_bits)` is an alias for `fixed.Shape(unsigned(i_bits + f_bits), f_bits)`.
-- `fixed.Shape.cast(shape, f_bits=0)`: Cast `shape` to a `fixed.Shape` instance.
 - `.i_bits`, `.f_bits`, `.signed`: Width and signedness properties of the `fixed.Shape`.
     - `.i_bits` includes the sign bit. That is, for `fixed.Shape(signed(16), 12)`, `.i_bits == 4`.
 - `.const(value)`: Create a `fixed.Const` from `value`.
@@ -67,7 +66,7 @@ The following operations are defined on it:
 - `.__add__(other)`, `.__radd__(other)`, `.__sub__(other)`, `.__rsub__(other)`, `.__mul__(other)`, `.__rmul__(other)`: Binary arithmetic operators.
   - If `other` is a `Value`, it'll be cast to a `fixed.Value` first.
   - If `other` is an `int`, it'll be cast to a `fixed.Const` first.
-  - If `other` is a `float`: TBD
+  - If `other` is a `float`, this is not permitted. The `float` must be explicitly cast to `fixed.Const`.
   - The result will be a new `fixed.Value` with enough precision to hold any resulting value without rounding or overflowing.
 - `.__lshift__(other)`, `.__rshift__(other)`: Bit shift operators.
     - `other` only accepts integral types. For example, shifting by a `float` or `fixed.Value` is not permitted.
@@ -128,6 +127,7 @@ TBD
     However, since a Python `float` is double precision, this means it's easy to make a >50 bit number by accident by doing something like `value * (1 / 3)`, and even if the result is rounded or truncated afterwards, the lower bits can affect rounding and thus won't be optimized out in synthesis.
   - We could use the same width for `other` as for `self`, adjusted to the appropriate exponent for the value.
   - We could outright reject it, requiring the user to explicitly specify precision like e.g. `value * Q(15).const(1 / 3)`.
+  - vk2seb@: I would lean toward outright rejecting this, with an explicit cast necessary (now reflected above).
 
 - How should we handle rounding?
   - Truncating and adding the most significant truncated bit is cheap and is effectively round to nearest with ties rounded towards positive infinity.
@@ -135,15 +135,21 @@ TBD
   - IEEE 754 defaults to round to nearest, ties to even, which is more expensive to implement.
   - Should we make it user selectable?
     - We still need a default mode used when a higher precision number is passed to `.eq()`.
+  - vk2seb@: Both truncation and simple rounding (round to nearest) are commonly used in DSP algorithms. For now, we provide only `truncate()` and `round()` strategies (now reflected above). Additional rounding strategies may be added in a future RFC, however we will always need a default rounding strategy.
 
 - Are there any other operations that would be good to have?
+  - From ld-cd@: `min()`, `max()` on `fixed.Shape` (vk2seb@: agree, heavily use this)
+  - From ld-cd@: `numerator()` as a signedness-preserving `as_value()` (vk2seb@: agree, heavily use this)
+  - vk2seb@: Let's add both of these operations (now reflected above)
 
 - Are there any operations that would be good to *not* have?
   - This API (`fixed.Shape.cast()`) seems confusing and difficult to use. Should we expose it at all? (@whitequark)
+  - vk2seb@: It can survive as a building block but I don't see why it needs to be exposed. Propose removal (reflected above).
 
 - `Decimal` and/or `Fraction` support?
   - This could make sense to have, but both can represent values that's not representable as binary fixed point.
     On the other hand, a Python `float` can perfectly represent any fixed point value up to a total width of 53 bits and any `float` value is perfectly representable as fixed point.
+  - vk2seb@: As both can represent values that aren't representable as binary fixed point, I don't see a huge benefit. I also can't immediately think of an application that would need >53 bit constants. I would propose for now we leave `Decimal` and `Fraction` out of scope of this RFC.
 
 - Name all the things.
   - Library name:
@@ -154,6 +160,12 @@ TBD
     - `.int_bits`, `.frac_bits`?
     - cursed option: `int, frac = x.width`?
   - `.round()` is a bit awkwardly named when it's used both to increase and decrease precision.
+  - vk2seb@: The existing modifications address this:
+    - Library name: `lib.fixed`
+    - Type names and shapes: signature has now been updated to use `i_bits`, `f_bits` and the explicit underlying storage in the constructor for `fixed.Shape`.
+    - We now have both `.round()` and `.truncate()`. I don't think using the same name for increasing and decreasing precision is so bad. But if you feel strongly about this we may consider:
+        - Renaming them.
+        - Disallowing increasing precision with these methods, and add a new method for precision extension .
 
 ## Future possibilities
 [future-possibilities]: #future-possibilities
--- a/text/0041-fixed-point.md
+++ b/text/0041-fixed-point.md
@@ -60,6 +60,9 @@ The following operations are defined on it:
   - If `value` is an `int` or `float`, it'll be cast to a `fixed.Const` first.
   - If `value` is a `fixed.Value`, the precision will be extended or rounded as required.
 - `.round(f_bits=0)`: Return a new `fixed.Value` with precision changed to `f_bits`, rounding as required.
+  - Rounding strategy: round to nearest with ties rounded towards positive infinity.
+  - Under the hood, this involves truncating and adding the most significant truncated bit.
+- `.truncate(f_bits=0)`: Return a new `fixed.Value` with precision changed to `f_bits`, truncating as required.
 - `.__add__(other)`, `.__radd__(other)`, `.__sub__(other)`, `.__rsub__(other)`, `.__mul__(other)`, `.__rmul__(other)`: Binary arithmetic operators.
   - If `other` is a `Value`, it'll be cast to a `fixed.Value` first.
   - If `other` is an `int`, it'll be cast to a `fixed.Const` first.
--- a/text/0041-fixed-point.md
+++ b/text/0041-fixed-point.md
@@ -1,0 +1,104 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [amaranth-lang/rfcs#41](https://github.com/amaranth-lang/rfcs/pull/41)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# Fixed point types
+
+## Summary
+[summary]: #summary
+
+Add fixed point types to Amaranth.
+
+## Motivation
+[motivation]: #motivation
+
+Fractional values in hardware are usually represented as some form of fixed point value.
+Without a first class fixed point type, the user has to manually manage how the value needs to be shifted to be represented as an integer and keep track of how that interacts with arithmetic operations.
+
+A fixed point type would encode and keep track of the precision through arithmetic operations, as well as provide standard operations for converting values to and from fixed point representations with correct rounding.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+TODO
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This RFC proposes a library addition `amaranth.lib.fixedpoint` with the following contents:
+
+`FixedPoint` is a `ShapeCastable` subclass.
+The following operations are defined on it:
+
+- `FixedPoint(f_width, /, *, signed)`: Create a `FixedPoint` with `f_width` fractional bits.
+- `FixedPoint(i_width, f_width, /, *, signed)`: Create a `FixedPoint` with `i_width` integer bits and `f_width` fractional bits.
+- `FixedPoint.cast(shape)`: Cast `shape` to a `FixedPoint` instance.
+- `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
+- `.const(value)`: Create a fixed point constant from an `int` or `float`, rounded to the closest representable value.
+- `.as_shape()`: Return the underlying `Shape`.
+- `.__call__(target)`: Create a `FixedPointValue` over `target`.
+
+`Q(*args)` is an alias for `FixedPoint(*args, signed=True)`.
+
+`UQ(*args)` is an alias for `FixedPoint(*args, signed=False)`.
+
+`FixedPointValue` is a `ValueCastable` subclass.
+The following operations are defined on it:
+
+- `FixedPointValue(shape, target)`: Create a `FixedPointValue` with `shape` over `target`.
+- `FixedPointValue.cast(value)`: Cast `value` to a `FixedPointValue`.
+- `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
+- `.shape()`: Return the `FixedPoint` this was created from.
+- `.as_value()`: Return the underlying value.
+- `.eq(value)`: Assign `value`.
+  - If `value` is a `FixedPointValue`, the precision will be extended or rounded as required.
+  - If `value` is an `int` or `float`, the value will be rounded to the closest representable value.
+  - If `value` is a `Value`, it'll be assigned directly to the underlying `Value`.
+- `.round(f_width=0)`: Return a new `FixedPointValue` with precision changed to `f_width`, rounding as required.
+- `.__add__(other)`, `.__radd__(other)`, `.__sub__(other)`, `.__rsub__(other)`, `.__mul__(other)`, `.__rmul__(other)`: Binary arithmetic operations.
+  - If `other` is a `Value` or an `int`, it'll be cast to a `FixedPointValue` first.
+  - If `other` is a `float`: TBD
+  - The result will be a new `FixedPointValue` with enough precision to hold any resulting value without rounding or overflowing.
+- `.__neg__()`, `.__pos__()`, `.__abs__()`: Unary arithmetic operations.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+TBD
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- `FixedPointValue.eq()` could cast a `Value` to a `FixedPointValue` first, and thereby shift the value to the integer part instead of assigning directly to the underlying value.
+  However, `Value.eq()` would always grab the underlying value of a `FixedPointValue`, and for consistency both `.eq()` operations should behave in the same manner.
+  - If we wanted to go the other way, this RFC could be deferred until another RFC allowing `ValueCastable` to override reflected `.eq()` have been merged.
+    However, having to explicitly do `value.eq(fp_value.round())` when rounding is desired is arguably preferable to having `value.eq(fp_value)` do an implicit rounding.
+
+- Unlike `.eq()`, it makes sense for arithmetic operations to cast a `Value` to `FixedPointValue`.
+  Multiplying an integer with a fixedpoint constant and rounding the result back to an integer is a reasonable and likely common thing to want to do.
+
+## Prior art
+[prior-art]: #prior-art
+
+[Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) is a common and convenient way to specify floating point types.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- What should we do if a `float` is passed as `other` to an arithmetic operation?
+  - We could use `float.as_integer_ratio()` to derive a perfect fixed point representation.
+    However, since a Python `float` is double precision, this means it's easy to make a >50 bit number by accident by doing something like `value * (1 / 3)`, and even if the result is rounded or truncated afterwards, the lower bits can affect rounding and thus won't be optimized out in synthesis.
+  - We could use the same width for `other` as for `self`, adjusted to the appropriate exponent for the value.
+  - We could outright reject it, requiring the user to explicitly specify precision like e.g. `value * Q(15).const(1 / 3)`.
+
+- There's two slightly different [Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) definitions, namely whether the bit counts includes the sign bit or not.
+  `UQ(15)` and `UQ(7, 8)` would be 15 bits in either convention, but `Q(15)` and `Q(7, 8)` would be 15 or 16 bits depending on the convention. Which do we pick?
+
+- Are there any other operations that would be good to have?
+
+- Bikeshed all the names.
+
+## Future possibilities
+[future-possibilities]: #future-possibilities
+
+TBD
--- a/text/0041-fixed-point.md
+++ b/text/0041-fixed-point.md
@@ -27,35 +27,38 @@ TODO
 
 This RFC proposes a library addition `amaranth.lib.fixed` with the following contents:
 
+### `fixed.Shape`
+
 `fixed.Shape` is a `ShapeCastable` subclass.
 The following operations are defined on it:
 
-- `fixed.Shape(f_width, /, *, signed)`: Create a `fixed.Shape` with zero integer bits and `f_width` fractional bits.
-- `fixed.Shape(i_width, f_width, /, *, signed)`: Create a `fixed.Shape` with `i_width` integer bits and `f_width` fractional bits.
-  - The sign bit is not included in `i_width` or `f_width`, so a `fixed.Shape(7, 8, signed=True)` will be 16 bits wide.
-- `fixed.Shape.cast(shape, f_width=0)`: Cast `shape` to a `fixed.Shape` instance.
-- `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
+- `fixed.Shape(shape, f_bits)`: Create a `fixed.Shape` with underlying storage `shape` and `f_bits` fractional bits.
+- The signedness is inherited from `shape`, so a `fixed.Shape(signed(16), 12)` would be a signed fixed-point number, 16 bits wide with 12 fractional bits.
+- A `fixed.Shape` may be constructed using the following aliases:
+    - `SQ(i_bits, f_bits)` is an alias for `fixed.Shape(signed(i_bits + f_bits), f_bits)`.
+    - `UQ(i_bits, f_bits)` is an alias for `fixed.Shape(unsigned(i_bits + f_bits), f_bits)`.
+- `fixed.Shape.cast(shape, f_bits=0)`: Cast `shape` to a `fixed.Shape` instance.
+- `.i_bits`, `.f_bits`, `.signed`: Width and signedness properties of the `fixed.Shape`.
+    - `.i_bits` includes the sign bit. That is, for `fixed.Shape(signed(16), 12)`, `.i_bits == 4`.
 - `.const(value)`: Create a `fixed.Const` from `value`.
 - `.as_shape()`: Return the underlying `Shape`.
 - `.__call__(target)`: Create a `fixed.Value` over `target`.
 
-`SQ(*args)` is an alias for `fixed.Shape(*args, signed=True)`.
-
-`UQ(*args)` is an alias for `fixed.Shape(*args, signed=False)`.
+### `fixed.Value`
 
 `fixed.Value` is a `ValueCastable` subclass.
 The following operations are defined on it:
 
 - `fixed.Value(shape, target)`: Create a `fixed.Value` with `shape` over `target`.
-- `fixed.Value.cast(value, f_width=0)`: Cast `value` to a `fixed.Value`.
-- `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
+- `fixed.Value.cast(value, f_bits=0)`: Cast `value` to a `fixed.Value`.
+- `.i_bits`, `.f_bits`, `.signed`: Width and signedness properties.
 - `.shape()`: Return the `fixed.Shape` this was created from.
 - `.as_value()`: Return the underlying value.
 - `.eq(value)`: Assign `value`.
   - If `value` is a `Value`, it'll be assigned directly to the underlying `Value`.
   - If `value` is an `int` or `float`, it'll be cast to a `fixed.Const` first.
   - If `value` is a `fixed.Value`, the precision will be extended or rounded as required.
-- `.round(f_width=0)`: Return a new `fixed.Value` with precision changed to `f_width`, rounding as required.
+- `.round(f_bits=0)`: Return a new `fixed.Value` with precision changed to `f_bits`, rounding as required.
 - `.__add__(other)`, `.__radd__(other)`, `.__sub__(other)`, `.__rsub__(other)`, `.__mul__(other)`, `.__rmul__(other)`: Binary arithmetic operators.
   - If `other` is a `Value`, it'll be cast to a `fixed.Value` first.
   - If `other` is an `int`, it'll be cast to a `fixed.Const` first.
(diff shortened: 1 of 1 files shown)


text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This RFC proposes a library addition `amaranth.lib.fixedpoint` with the following contents:
+
+`FixedPoint` is a `ShapeCastable` subclass.
+The following operations are defined on it:
+
+- `FixedPoint(f_width, /, *, signed)`: Create a `FixedPoint` with `f_width` fractional bits.
+- `FixedPoint(i_width, f_width, /, *, signed)`: Create a `FixedPoint` with `i_width` integer bits and `f_width` fractional bits.
```
whitequark, line 34: These two lines seem mutually incompatible.

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+TODO
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This RFC proposes a library addition `amaranth.lib.fixedpoint` with the following contents:
```
whitequark, line 28: Bikeshed: `lib.fixed`, `lib.fixnum`.

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+
+`FixedPoint` is a `ShapeCastable` subclass.
+The following operations are defined on it:
+
+- `FixedPoint(f_width, /, *, signed)`: Create a `FixedPoint` with `f_width` fractional bits.
+- `FixedPoint(i_width, f_width, /, *, signed)`: Create a `FixedPoint` with `i_width` integer bits and `f_width` fractional bits.
+- `FixedPoint.cast(shape)`: Cast `shape` to a `FixedPoint` instance.
```
whitequark, line 35: What is the semantics of this operation?

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+- `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
+- `.const(value)`: Create a fixed point constant from an `int` or `float`, rounded to the closest representable value.
```
whitequark, line 37: `Decimal` support as well?

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+
+This RFC proposes a library addition `amaranth.lib.fixedpoint` with the following contents:
+
+`FixedPoint` is a `ShapeCastable` subclass.
+The following operations are defined on it:
+
+- `FixedPoint(f_width, /, *, signed)`: Create a `FixedPoint` with `f_width` fractional bits.
+- `FixedPoint(i_width, f_width, /, *, signed)`: Create a `FixedPoint` with `i_width` integer bits and `f_width` fractional bits.
+- `FixedPoint.cast(shape)`: Cast `shape` to a `FixedPoint` instance.
+- `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
```
whitequark, line 36: I feel like the `i_width` and `f_width` names are difficult enough to read that it's of more importance than bikeshedding to come up with something more readable.

`.int_bits`, `.frac_bits`?

cursed option: `int, frac = x.width`?

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+- `.const(value)`: Create a fixed point constant from an `int` or `float`, rounded to the closest representable value.
+- `.as_shape()`: Return the underlying `Shape`.
+- `.__call__(target)`: Create a `FixedPointValue` over `target`.
+
+`Q(*args)` is an alias for `FixedPoint(*args, signed=True)`.
```
whitequark, line 41: I understand Q-notation is an established one, however since Amaranth defaults to unsigned, I feel that having `SQ` and `UQ` (without having just `Q`) would serve Amaranth better.

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+The following operations are defined on it:
+
+- `FixedPointValue(shape, target)`: Create a `FixedPointValue` with `shape` over `target`.
+- `FixedPointValue.cast(value)`: Cast `value` to a `FixedPointValue`.
+- `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
+- `.shape()`: Return the `FixedPoint` this was created from.
+- `.as_value()`: Return the underlying value.
+- `.eq(value)`: Assign `value`.
+  - If `value` is a `FixedPointValue`, the precision will be extended or rounded as required.
+  - If `value` is an `int` or `float`, the value will be rounded to the closest representable value.
```
whitequark, line 55: What happens if you assign 1024 to a Q8.8 value?

whitequark (commented):
General comments:

- We designed `lib.data` and `lib.wiring` to be imported as entire modules: `from amaranth.lib import data, wiring`. This library should be designed (naming-wise) to work that way too. `fixed.Shape` and `fixed.Value` are one option, though I can see why others may object to it.
- I feel like implicit conversions from floats are potentially a source of bugs severe enough that maybe we shouldn't have them at all.
- How does one perform computation on fixed point values during compilation? I think "you just use floats" is an unsatisfying answer.

text/0041-fixed-point.md
zyp: The idea is that the underlying implementation would be `FixedPoint(i_or_f_width, f_width = None, /, *, signed)`, but I found it clearer to express it like that. Consider how you can do `range(stop)` and `range(start, stop)`.

text/0041-fixed-point.md
zyp: And probably `Fraction` as well.

text/0041-fixed-point.md
zyp: I just updated it to `.cast(shape, f_width=0)` locally. The idea is to determine `i_width` automatically, so `FixedPoint.cast(unsigned(8))` would result in `UQ(8)` and `FixedPoint.cast(unsigned(8), f_width = 4)` would result in `UQ(4, 4)`.

text/0041-fixed-point.md
whitequark: Makes sense. I wasn't sure if it was that or a typo. Does the first constructor add a sign bit when `signed` is true, and otherwise create a number whose range spans from 0 to some value below 1?

text/0041-fixed-point.md
whitequark: Why `UQ(4,4)`?

text/0041-fixed-point.md
zyp: Yes, although the exact semantics wrt. the sign bit depends on which Q notation we settle on.

text/0041-fixed-point.md
zyp: `FixedPoint.cast()` is a building block for `FixedPointValue.cast()`. The idea is to be able to say «here's a value with n fractional bits, please turn it into an appropriately sized `FixedPointValue`». An `unsigned(8)` with four fractional bits would have four integer bits left and thus cast to `UQ(4, 4)`.

text/0041-fixed-point.md
zyp: `Signal(Q(8, 8)).eq(1024)` would effectively be `Signal(signed(17)).eq(1024 << 8)`. (Or `signed(16)` depending on Q notation.)

Speaking of overflow, I figure fixed point overflow should behave like regular integer overflow. Saturating math feels orthogonal to this RFC and could be added through a later RFC that handles both integer and fixed point values.

zyp:
> I feel like implicit conversions from floats are potentially a source of bugs severe enough that maybe we shouldn't have them at all.

I think for making constants of a specific shape it's reasonable to have implicit float conversion. As the `other` argument to binary operators it's probably reasonable to require an explicit conversion to a constant first.

> * How does one perform computation on fixed point values during compilation? I think "you just use floats" is an unsatisfying answer.

An adjacent question is «what will `FixedPointValue.get()` in the simulator return?» We should probably add a separate `FixedPointConstant` type.

text/0041-fixed-point.md
whitequark: This API seems confusing and difficult to use. Should we expose it at all?

whitequark:
> We should probably add a separate `FixedPointConstant` type.

Are there good fixed point libraries for Python we could use here?

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+  - If we wanted to go the other way, this RFC could be deferred until another RFC allowing `ValueCastable` to override reflected `.eq()` have been merged.
+    However, having to explicitly do `value.eq(fp_value.round())` when rounding is desired is arguably preferable to having `value.eq(fp_value)` do an implicit rounding.
+
+- Unlike `.eq()`, it makes sense for arithmetic operations to cast a `Value` to `FixedPointValue`.
+  Multiplying an integer with a fixedpoint constant and rounding the result back to an integer is a reasonable and likely common thing to want to do.
+
+## Prior art
+[prior-art]: #prior-art
+
+[Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) is a common and convenient way to specify floating point types.
```
adamgreig, line 83:
```suggestion
[Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) is a common and convenient way to specify fixed point types.
```

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+[unresolved-questions]: #unresolved-questions
+
+- What should we do if a `float` is passed as `other` to an arithmetic operation?
+  - We could use `float.as_integer_ratio()` to derive a perfect fixed point representation.
+    However, since a Python `float` is double precision, this means it's easy to make a >50 bit number by accident by doing something like `value * (1 / 3)`, and even if the result is rounded or truncated afterwards, the lower bits can affect rounding and thus won't be optimized out in synthesis.
+  - We could use the same width for `other` as for `self`, adjusted to the appropriate exponent for the value.
+  - We could outright reject it, requiring the user to explicitly specify precision like e.g. `value * Q(15).const(1 / 3)`.
+
+- There's two slightly different [Q notation](https://en.wikipedia.org/wiki/Q_(number_format)) definitions, namely whether the bit counts includes the sign bit or not.
+  `UQ(15)` and `UQ(7, 8)` would be 15 bits in either convention, but `Q(15)` and `Q(7, 8)` would be 15 or 16 bits depending on the convention. Which do we pick?
```
adamgreig, line 95: In practice I'm not sure how common the "ARM" i.e. "include sign bit in n" convention is. Even Arm don't use it for modern software, e.g. "Q15" in [CMSIS-DSP](https://arm-software.github.io/CMSIS-DSP/latest/structarm__bilinear__interp__instance__q15.html). If we do have `Q(n)`, I think it should be `Q(15)` to mean a 16-bit signed value. However, if we did go with @whitequark's suggestion of `SQ(n)`, perhaps we have some more room to manoeuvre?

My preference would still be to omit the sign bit from the count, which presumably also matches the arguments to the `FixedPoint(f_width, /, *, signed)` constructor? In other words, I assumed `FixedPoint(15, signed=True)` creates a 16-bit `Value`, since it has 15 "fractional bits" which aren't sign bits?

text/0041-fixed-point.md
```diff
@@ -0,0 +1,104 @@
...
+
+- Are there any other operations that would be good to have?
```
adamgreig, line 97: Perhaps `__rshift__`, `__rrshift__`, `__lshift__`, `__rlshift__`?

text/0041-fixed-point.md
zyp: Yes, `Q(15)` or `SQ(15)` would be an alias for `FixedPoint(15, signed=True)` whichever way we decide to go.

text/0041-fixed-point.md
zyp: Agreed, `__rshift__` and `__lshift__` sounds reasonable. The reflected variants are probably not useful since it doesn't make sense to shift anything by a non-integer value.

text/0041-fixed-point.md
```diff
@@ -91,12 +110,23 @@ TBD
...
-
 - Are there any other operations that would be good to have?
 
-- Bikeshed all the names.
+- Are there any operations that would be good to *not* have?
+  - This API (`fixed.Shape.cast()`) seems confusing and difficult to use. Should we expose it at all? (@whitequark)
+
+- `Decimal` and/or `Fraction` support?
+  - This could make sense to have, but both can represent values that's not representable as binary fixed point.
+    On the other hand, a Python `float` can perfectly represent any fixed point value up to a total width of 53 bits and any `float` value is perfectly representable as fixed point.
```
whitequark, line 120: OK, this is a compelling argument for support of floats. I am convinced that it is a good idea to have floats supported in at least some way. But we need to be very careful around overflows.

text/0041-fixed-point.md
```diff
@@ -0,0 +1,143 @@
...
+    However, since a Python `float` is double precision, this means it's easy to make a >50 bit number by accident by doing something like `value * (1 / 3)`, and even if the result is rounded or truncated afterwards, the lower bits can affect rounding and thus won't be optimized out in synthesis.
+  - We could use the same width for `other` as for `self`, adjusted to the appropriate exponent for the value.
+  - We could outright reject it, requiring the user to explicitly specify precision like e.g. `value * Q(15).const(1 / 3)`.
+
+- How should we handle rounding?
+  - Truncating and adding the most significant truncated bit is cheap and is effectively round to nearest with ties rounded towards positive infinity.
+  - Simply truncating is free, rounds towards negative infinity.
+  - IEEE 754 defaults to round to nearest, ties to even, which is more expensive to implement.
+  - Should we make it user selectable?
+    - We still need a default mode used when a higher precision number is passed to `.eq()`.
```
whitequark, line 119: I wonder if we could leave it unspecified and implementation-defined until we can get a numerics expert to chime in.

text/0041-fixed-point.md
```diff
@@ -0,0 +1,143 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+This RFC proposes a library addition `amaranth.lib.fixed` with the following contents:
+
+`fixed.Shape` is a `ShapeCastable` subclass.
+The following operations are defined on it:
+
+- `fixed.Shape(f_width, /, *, signed)`: Create a `fixed.Shape` with zero integer bits and `f_width` fractional bits.
+- `fixed.Shape(i_width, f_width, /, *, signed)`: Create a `fixed.Shape` with `i_width` integer bits and `f_width` fractional bits.
+  - The sign bit is not included in `i_width` or `f_width`, so a `fixed.Shape(7, 8, signed=True)` will be 16 bits wide.
```
whitequark, line 35: This means that `fixed.Shape(7, 0, signed=True)` has the same width as `signed(8)`, which seems counterintuitive.

text/0041-fixed-point.md
```diff
@@ -0,0 +1,143 @@
...
+
+`fixed.Const` is a `fixed.Value` subclass.
+The following additional operations are defined on it:
+
+- `fixed.Const(value, shape=None)`: Create a `fixed.Const` from `value`. `shape` must be a `fixed.Shape` if specified.
+  - If `value` is an `int` and `shape` is not specified, the smallest shape that will fit `value` will be selected.
+  - If `value` is a `float` and `shape` is not specified, the smallest shape that gives a perfect representation will be selected.
+    If `shape` is specified, `value` will be rounded to the closest representable value first.
+- `.as_integer_ratio()`: Return the value represented as an integer ratio `tuple`.
+- `.as_float()`: Return the value represented as a `float`.
```
whitequark, line 76: What happens if it's not representable? At least we need to have `c.as_float(exact=True)` which gives an error in that case.

text/0041-fixed-point.md
```diff
@@ -0,0 +1,143 @@
...
+- Operators are extended to return a `fixed.Const` if all operands are constant.
```
whitequark, line 77: This is inconsistent with how `Const()` works.

text/0041-fixed-point.md
samimia-swks: In most DSP applications, simple truncating is done (bit picking, which is equivalent to a floor()) because it's free. I would vote for that being the default behavior at least.

text/0041-fixed-point.md
samimia-swks: The dominant notation in the industry, at least in the audio ASIC world, is to include the sign bit in the number of integer bits. For example, -1 to 1 is represented as a Q1.23. That would be my vote.

text/0041-fixed-point.md
samimia-swks: +1 on `int_bits`, `frac_bits`...

cr1901:
I have nothing concrete here, just minddumping stuff I wrote down in text files locally.

I assume that `__div__` is deliberately not implemented. Thanks to the curse of rational numbers you'd need infinite bits to represent many divisions without loss, such as 1/5. Amaranth extends width so values can be completely represented, but this is impossible for fixed point division.

You can get as much precision as you'd like from dividing fixed point numbers by left-shifting a numerator (increasing its `f_width`) before dividing. Consider dividing e.g. 1*10^-6 by 999999*10^-6:

* `1*10^-6/999999*10^-6 = int(1/999999) * (10^-6/10^-6) = 0*10^0`.
  *  Q(0,6)/Q(0,6) yields Q(6,0), yields 0 due to not enough precision.

On the other hand, we can left-shift the integer part to get a non-zero result from this divide. :

* `1*10^-6/999999*10^-6 = 1*10^-6*(10^7*10^-7)/999999*10^-6  = int(10000000/999999)*(10^-13/10^-6) = 10*10^-7`
  * Q(0,13)/Q(0,6) yields Q(6,7), yields a non-zero result (close enough).

I arbitrarily chose to shift by 7 decimal places (replace with binary for this RFC), but it's not obvious to me that any given left shift will be universally good for all use cases to prevent divides from returning 0 for non-zero numerators.

Additionally, you may only want the extra precision during the divide; e.g. I could truncate the Q(6,7) in the second example back down to Q(0,7), saving 6 bits, and still have a perfectly valid result to pipe to the rest of a design. Whether/how much to truncate depends on the dynamic range of fixed point values you expect your design to see.

Will fixed point divide be rare enough that divide should be completely ignored? Maybe there should be a divide function that allows the user to tweak:

* Extra precision used during the integer divide portion.
* The `Q` of the output.

I do not have a concrete use case in mind right now (I could probably shoehorn division into my Celsius to Fahrenheit conversion experiments). Just something I mulled over today.

whitequark:
You could always multiply by reciprocal for an application like unit conversion.

cr1901:
Multiply-by-reciprocal only works if you can calculate it ahead of time- i.e. one of the multiply inputs is constant. Otherwise you'd have to to find the reciprocal while your design runs before doing the multiply, which... requires a divide. The thing you're trying to avoid.

cr1901:
Indeed, I will need to find a different way to shoehorn a division into my Celsius to Fahrenheit experiments :).

Divide by non-constant factor for fixed point is probably uncommon. But I hesitate to say it's "so uncommon it's not worth supporting in some way at all, even if a `__div__` impl is impossible".

zyp:
The question of whether `__div__` makes sense to have is whether it's feasible for synthesis to infer a divider.

Independent of that, there's already nothing that prevents you from making a divider component with two fixedpoint inputs and a fixedpoint output.

cr1901:
> The question of whether __div__ makes sense to have is whether it's feasible for synthesis to infer a divider.

Do FPGAs have divider IPs? I thought they only had multiplier IP. I assumed amaranth [provides](https://amaranth-lang.org/docs/amaranth/latest/guide.html#arithmetic-operators) floor division/mod for simulation purposes, and not something you'd want to synthesize.

whitequark:
> is whether it's feasible for synthesis to infer a divider.

Yosys will actually synthesize a divider for you if you do `a // b` in Amaranth. Here's the size for a 8/8=8 combinational divider for iCE40:

```
   Number of cells:                114
     SB_CARRY                       64
     SB_LUT4                        50
```

cr1901:
> The question of whether __div__ makes sense to have

Even if dividers synthesize, I don't think `__div__` makes sense to have for fixed point; in integer division, "`d` doesn't divide `n` evenly" is handled by remainder. Therefore all values can be represented in the output without loss, which is great since Amaranth [never overflows](https://amaranth-lang.org/docs/amaranth/latest/guide.html#arithmetic-operators).

In fixed point, there's no remainder, so if  "`d` doesn't evenly divide `n`"  we keep adding bits to `n` until `d` evenly divides our modified `n`. This can require infinite number of extra added bits, and if we keep with "Amaranth arithmetic never overflows" semantics, this is impossible to satisfy.

Even if `__div__` doesn't make sense, I think there should be _something_ for divides (a function instead of an overloaded operator?) [Last night](https://github.com/amaranth-lang/rfcs/pull/41#issuecomment-1965625191) was me minddumping my thoughts on what fixed point divide should handle.

zyp:
If a perfect `__div__` is desirable to have, instead of a `fixed.Value`, we could simply have it return a ratio object containing the two operands. Once such a ratio is passed to e.g. `fixed.Value.eq()`, the actual division could be performed with the precision of the assignment target.

To avoid scope creep, I'm inclined to leave inferred division out of this RFC. We could instead do a separate RFC later for that.

cr1901:
> To avoid scope creep, I'm inclined to leave inferred division out of this RFC. We could instead do a separate RFC later for that.

That's fine, mind putting your comment re: ratio objects under Future Possibilities/making a mention that division is out of scope?

text/0041-fixed-point.md
ld-cd: Truncating can also be a big foot gun in some DSP applications that is hard to track down, you can end up with a DC spur that grows over time or over your pipeline with no obvious explanation. A middle ground is round towards zero or symmetrically towards positive/negative infinity (this is nearly free on the output of Xilinx DSP48E1/2 and I believe cheap/free on ECP5 and friends DSPs). The way I personally would handle it would be to make the rounding mode part of the `FixedPoint` type signature making it `FixedPoint(i_width, f_width, /, *, signed, rounding_mode)`. Truncate is still a reasonable default for most applications though.

text/0041-fixed-point.md
ld-cd, line 35: Is having `i_width` positive and `f_width` negative or the converse in scope for this RFC? If it is then I think `.round` should take `i_width` as an optional argument

whitequark: What would that mean?

text/0041-fixed-point.md
whitequark: Hm, making `rounding_mode` part of the signature is a pretty big action, what happens if you add two numbers with differing rounding modes for example?

text/0041-fixed-point.md
ld-cd: Two scenarios where I think you could reasonably end up in this position with the current RFC

1. You start off with say UQ0.16 [0,1) and right shift by 2 the natural representation is still 16 bits but with the range [0, 0.25) and I believe the representation is going to end up being UQ-2.18 in this notation?
2. This is less reasonable but say you throw a 4096 (1<<12) point FFT at values with the format SQ7.10, if you are naive about the scaling and keep the bit width to 18 bits, you end up with SQ19.-2 (ie a step of 4 between every point). This is equivalent in terms of the bits on the wire to starting with SQ0.17 and ending with SQ12.5 but it feels more silly and you can end up there naturally so it should probably be supported

I'm currently working on tracking down some DSP bugs in a project at work that nobody every simulated, in my fixed point emulation code the number format I chose was basically n_bits, exponent which handles these cases more naturally but is probably overall less intelligible. I'd imagine this looking like `fixed.Shape(shape, exponent)` in amaranth (IE `fixed.Shape(signed(18), 2)` for SQ19.-2) but that would be a pretty large change I'm not sure it would be a net positive anyways (also don't want to bikeshed here).

text/0041-fixed-point.md
ld-cd: To be clear the semantics I'm imagining here are if you have `a` and `b` and `b` has more fractional bits than `a` then when you assign `a.eq(b)` the rounding mode that is used is that of `a` so that downstream DSP modules can enforce a rounding mode on their inputs.

The scenarios where I can imagine what you propose causing issues are:

1. User adds `a` and `b` and then `.round`s them
2. A user adds `a` and `b` and utilizes the result as a port on their module.

Solutions that solve 1. But not two in no particular order:

1. Don't actually resolve the round until a signal is assigned to something with a concrete value (this seems like a bad idea and probably doesn't solve the problem)
2. Make the rounding mode undefined and require an explicit rounding mode to `.round`, but this doesn't solve the problem of getting a top level port with an undefined rounding mode

Solutions that I believe would solve both:

1. Choose the rounding mode of the left argument
2. Choose the less footgunny rounding mode (TOEVEN,TOODD>SYMZERO,SYMINF>TRUNC) with ties going to either a defined order or the left most
3. Ban arithmetic options between numbers with different rounding modes and require an explicit cast of one of the arguments with something like `.with_mode(other.rounding_mode)`
4. Make this undefined behavior and pick one of the above for now

My naive preference inclination would be 3, it seems like a fairly rare scenario and if you end up in it it's probably worth making the user think about what they want the result to be, but I'm not a language designer so take that with a grain of salt

ld-cd: Further thoughts on 3 as an option:

Operations with a constant fixed point value should probably inherit the rounding mode of the non-constant value without requiring an explicit cast because I think the alternative is annoying to deal with (requiring a shape to be passed any time a constant is declared).

text/0041-fixed-point.md
ld-cd: Ok I've played with what the implementation does here, this might be a bit long; The implementation does allow the creation of the above types of representations (IE Q-4.20, Q20.-4) and appears to operate on them correctly, however it for the most part won't generate them on its own. Left shift and right shift preserve the number of bits until either `i_width` or `f_width` would go negative at which point it pads things out:

Right shift works as expected:
```python
In [2]: fixed.SQ(8, 8), fixed.SQ(-4, 20), Signal(fixed.SQ(8, 8)) >> 8, Signal(fixed.SQ(8, 8)) >> 12
Out[2]:
(fixed.Shape(8, 8, signed=True),
 fixed.Shape(-4, 20, signed=True),
 (fixedpoint SQ0.16 (sig $signal)),
 (fixedpoint SQ0.20 (sig $signal)))
```

Left shift seems to get converted to unsigned which I think is not the correct behavior:
```python
In [4]: Signal(fixed.SQ(8, 8)) << 8, Signal(fixed.SQ(8, 8)) << 9, Signal(fixed.SQ(8, 8)) << 10, Signal(fixed.SQ(8, 8)) << 12
Out[4]:
((fixedpoint SQ16.0 (sig $signal)),
 (fixedpoint UQ18.0 (cat (const 1'd0) (sig $signal))),
 (fixedpoint UQ19.0 (cat (const 2'd0) (sig $signal))),
 (fixedpoint UQ21.0 (cat (const 4'd0) (sig $signal))))
```

I personally think the rounding semantics are a bit difficult to reason about as they stand:
```python
In [5]: (Signal(fixed.SQ(8, 8)) << 12).round(-4), (Signal(fixed.SQ(8, 8)) >> 12).round(16)
Out[5]:
((fixedpoint UQ21.-4 (+ (slice (cat (const 4'd0) (sig $signal)) 4:21) (slice (cat (const 4'd0) (sig $signal)) 3:4))),
 (fixedpoint SQ0.16 (+ (slice (sig $signal) 4:17) (slice (sig $signal) 3:4))))
```
Rounding with negative fractional bits produces the sensible result (wrt signed to unsigned bug), but rounding off the fractional bits in a right shift looses precision when it does not necessarily need too. I think this is a decent argument for either always preserving the number of bits in a shift, or including an `i_width` argument to round. Personally I would advocate for always preserving the number of bits because I think that that is easier to reason about and the argument to round can just be `width` not `i_width` or `f_width`.

Addition between a number with no fractional bits and one with no integer bits also produces a number 2 bits wider than I believe is necessary:

```python
In [7]: Signal(fixed.SQ(-4, 20)) + Signal(fixed.SQ(20, -4))
Out[7]: (fixedpoint SQ22.20 (+ (sig $signal) (cat (const 24'd0) (sig $signal))))
```

text/0041-fixed-point.md
```diff
@@ -0,0 +1,143 @@
...
+
+`fixed.Shape` is a `ShapeCastable` subclass.
+The following operations are defined on it:
+
+- `fixed.Shape(f_width, /, *, signed)`: Create a `fixed.Shape` with zero integer bits and `f_width` fractional bits.
+- `fixed.Shape(i_width, f_width, /, *, signed)`: Create a `fixed.Shape` with `i_width` integer bits and `f_width` fractional bits.
+  - The sign bit is not included in `i_width` or `f_width`, so a `fixed.Shape(7, 8, signed=True)` will be 16 bits wide.
+- `fixed.Shape.cast(shape, f_width=0)`: Cast `shape` to a `fixed.Shape` instance.
+- `.i_width`, `.f_width`, `.signed`: Width and signedness properties.
+- `.const(value)`: Create a `fixed.Const` from `value`.
```
ld-cd, line 38: would a `.max()` and `.min()` method or property make sense here to get a const with the max/min representable value make sense here or is that outside the scope of the `Shape` API given that the base signed and unsigned types don't have that?

text/0041-fixed-point.md
ld-cd: I just managed to hit this myself even though I've read through this a few times

zyp: I've thought about this since I wrote the last draft and concluded that this is probably the wrong decision, so I'm intending to change it when I find time to revise the RFC again.

text/0041-fixed-point.md
```diff
@@ -0,0 +1,143 @@
...
+  - If `value` is a `Value`, it'll be assigned directly to the underlying `Value`.
+  - If `value` is an `int` or `float`, it'll be cast to a `fixed.Const` first.
+  - If `value` is a `fixed.Value`, the precision will be extended or rounded as required.
+- `.round(f_width=0)`: Return a new `fixed.Value` with precision changed to `f_width`, rounding as required.
+- `.__add__(other)`, `.__radd__(other)`, `.__sub__(other)`, `.__rsub__(other)`, `.__mul__(other)`, `.__rmul__(other)`: Binary arithmetic operators.
+  - If `other` is a `Value`, it'll be cast to a `fixed.Value` first.
+  - If `other` is an `int`, it'll be cast to a `fixed.Const` first.
+  - If `other` is a `float`: TBD
+  - The result will be a new `fixed.Value` with enough precision to hold any resulting value without rounding or overflowing.
+- `.__lshift__(other)`, `.__rshift__(other)`: Bit shift operators.
```
ld-cd, line 64: In the implementation at least the semantics of these seem to be different from underlying amaranth values. If the shift amount is an integer these act like `.shift_left()` and `.shift_right()`. Those operators should probably be added here and the semantics of `__lshift__` and `__rshift__` adjusted to match those on `unsigned()` and `signed()`.

Note: I may be wrong here, still learning the language

text/0041-fixed-point.md
```diff
@@ -0,0 +1,143 @@
...
+- `.__neg__()`, `.__pos__()`, `.__abs__()`: Unary arithmetic operators.
+- `.__lt__(other)`, `.__le__(other)`, `.__eq__(other)`, `.__ne__(other)`, `.__gt__(other)`, `.__ge__(other)`: Comparison operators.
```
ld-cd, line 66: For the comparison operators the rounding behavior seems under defined when the types don't have the same precision, the options seem to be:

1. Round to the precision of the left argument (what the draft implementation does)
2. Round to the precision of the least precise argument (easiest for the hardware)
3. Round to the most precise argument (excluding constants maybe?)
4. Ban comparison between values (but maybe not const and value?) with different precision

text/0041-fixed-point.md
```diff
@@ -0,0 +1,143 @@
...
+  - We could outright reject it, requiring the user to explicitly specify precision like e.g. `value * Q(15).const(1 / 3)`.
+
+- How should we handle rounding?
+  - Truncating and adding the most significant truncated bit is cheap and is effectively round to nearest with ties rounded towards positive infinity.
+  - Simply truncating is free, rounds towards negative infinity.
+  - IEEE 754 defaults to round to nearest, ties to even, which is more expensive to implement.
+  - Should we make it user selectable?
+    - We still need a default mode used when a higher precision number is passed to `.eq()`.
+
+- Are there any other operations that would be good to have?
```
ld-cd, line 121: `.as_value()` returns a value with a signdedness that doesn't depend on the signdedness of the fixedpoint shape, should there be a `.lower()` or `.numerator()` method that returns `.as_value()` cast to the appropriate signdedness?

text/0041-fixed-point.md
ld-cd: 4. seems bad, I'm going to implement 3 for my FFT testing and see if it ends up being a problem

jrmoserbaltimore:
Is this still being worked on?

I'm familiar with the Python FxpMath module and so like the idea of denoting Qs.m or SQs.m for significand and mantissa signed, UQs.m for unsigned. The syntax @ld-cd proposed looks good to me.  I'm not so into the `fixed.Shape(f_width, ...)` and `fixed.Shape(i_width, f_width, ...)` notably because the first parameter means something different depending on if the first two are integers (wat?); but I think it's more clear to have something like SQ() and UQ() requiring both `s` and `m`.  I also like the syntax `SQ(5, -6)` for shifting left (xxxxx000000) and `SQ(-4,5)` for shifting right (+.+++xxxxx), note that `SQ(-1,m)` puts the MSB at the 1/2 (0.1) position. 

It looks like there's still question about what to do if the other value is a `float`.  I assume this means if a Python `float` or `numpy` floating-point type is passed.  I like the explicit conversion approach.  I'm not keen on spending a whole bunch of time trying to debug something that isn't behaving the way I expected, or wondering if it's going to behave the way I expect instead of blindly hoping.

The same can be said for passing any constant as a fixed-point value:  I'd rather require the constant to be presented as `SQ()` or `UQ()` rather than "here's a Python type, have a guess at what's the right thing to do."

In the worst case, you can always add implicit conversion later without it being a breaking change.

**Use case**:  I frequently use lookup tables generated by `numpy`.  These tables may be floats.  If I have a table of `np.float64` or such I want to be able to somehow indicate that the table is a certain fixed-point type.  To my mind, this is pretty simple:  I tell it I want a `SQ()` or `UQ()` somehow (e.g. `SQ(-4, 16, log2_table)`) and Amaranth somehow does the right thing (which ultimately means converting the entire addressable portion of the table and embedding it in the design).  That's a relatively simple problem.

I'm using an Abed-Siferd predictor to produce an accurate `log2()` and `exp2()` down to more than 4 bits with simple circuitry; I need 19 bits of mantissa.  My solution is a 512-entry `SQ(-4,16)` look-up table added to the `SQ(4.5)` output of the predictor.  Linear interpolation between the resulting values supplies further granularity.

To do this, I need to add dissimilar fixed-point values with proper sign extension:  adding a negative `SQ(-4,16`) to a `Q(12,4)` needs to extend all the sign bits to cover the `Q(12,4)` (whether SQ or UQ).  The same goes for aligning outputs.

I'm looking forward to built-in fixed-point math, the lack of which is a deficiency in VHDL and Verilog that I would very much like to escape.

whitequark:
> Is this still being worked on?

We're definitely still very interested in it!

The RFC will need several people driving it forward; usually two is enough (I am usually one of them and the author is another) but this one will likely need three. At the moment I'm seriously ill and will probably continue to be mostly bedridden until January (chronic pain that became worse recently).

I think it's definitely worth it to work out a bunch of kinks before then.

zyp:
> Is this still being worked on?

On my side this is currently shelved for a lack of time; I have higher priority projects than the one I need fixedpoint for, but I intend to eventually come back to it.

If you (or anybody else) would like to have a go at revising/improving the RFC instead of waiting for me to find time for it, I welcome you.

jrmoserbaltimore:
Fair enough, life happens; just wanted to weigh in though, this is seriously useful to me in basically anything involving signal processing.

whitequark:
Yep, I think it's incredibly important to have. I have my own interest in the RFC too.

schnommus:
I don't want to spam this thread with +1s, but as a [heavy user](https://github.com/apfelaudio/tiliqua/blob/main/gateware/src/tiliqua/dsp.py) of a modified version of this RFC already, I'm also happy to help out in any way I can.

whitequark:
@schnommus Could you list your modifications?

schnommus:
@whitequark I have tried to keep things in line with this draft and @zyp's original PR to reduce the risk of churn, so there is not so much that deviates. In general I like the RFC in its current form. Some small things I changed:

- **`min()`, `max()` representable constants from a shape:** this is a +1 to @ld-cd 's comment on having `min()`, `max()` methods for retrieving the span of representable values for a particular shape. It has proven quite useful.
- **behavior of `fixed.Const` initialization** should be specified for cases where a shape is specified, but the provided `int` or `float` is out of range of that shape. An exception may be a reasonable default, however clamping a float initializer to the max- or min- possible representable value within that shape also seems often desirable. Perhaps as an optional switch that demotes such exceptions to a warning. The alternative is to clamp the float before using it as an initializer, but this gets clunky if you want it to clamp to the true max/min representable values.
- **dedicated `truncate()` method alongside the existing `round()` method**: this could be collapsed into the same `round()` function with different arguments, but I found it less noisy to read DSP code when there are dedicated and separate `truncate()` and `round()` methods. It's pretty clear what `truncate` means, but `round` can have different semantics.
- **`.eq(value)` and other ops should preserve `signed` nature of the LHS `fixed` even if the underlying `Value` is `unsigned`**. This is a problem as `as_value()` does not necessarily retain the signedness of the underlying fixed.Shape. This was a practical matter I ran into that is not really specified here. But in general the RFC should be careful to specify how signedness affects each operation, whether the operands are simple `Value`s or `fixed.Value`s.

That's all I touched. On some other bikeshedding points from above:
- +1 on `int_bits` / `frac_bits` rather than `i_width`, `f_width`.
- +1 that `fixed.Shape(7, 0, signed=True)` counterintuitively has the same width as `signed(8)`. Would be nice to find another solution
- A few of the operators need more detailed specification e.g.:
  - What should `fixed.Value() >> fixed.Value()` mean? Shift by the rounded integral component or throw an exception?
  - What are the rounding implications of performing `__eq__` or `__lt__` on `fixed.Value`s of different widths? @ld-cd also mentioned this above

samimia-swks:
In case this is useful, I have developed a fixed-point library for systemverilog which we have open-sourced here: https://github.com/SkyworksSolutionsInc/fplib

The conventions used in this library are wide-spread at least in the audio ASIC industry.

text/0041-fixed-point.md
whitequark: We could definitely have additional functionality on specific shapes, e.g. enum shapes are pretty magical, so are layouts. `.min()`/`.max()` seem completely reasonable to me.

whitequark:
@schnommus These all look reasonable to me although I am not a DSP expert so I may well be missing something. From what @zyp says it might be meaningful for you to pick up the RFC yourself, what do you think?

schnommus:
> @schnommus These all look reasonable to me although I am not a DSP expert so I may well be missing something. From what @zyp says it might be meaningful for you to pick up the RFC yourself, what do you think?

@whitequark sure, I'm happy to take this. As a first action I would try to condense all comments above into the existing markdown from @zyp . Ideally we keep the changes in this PR to retain the comment history.

I can't commit to a hard deadline due to work commitments, but my personal goal would be to get a new RFC revision out in the coming week.

zyp:
> * +1 on `int_bits` / `frac_bits` rather than `i_width`, `f_width`.
> * +1 that `fixed.Shape(7, 0, signed=True)` counterintuitively has the same width as `signed(8)`. Would be nice to find another solution

I propose that instead of specifying the integer and fractional parts individually, we specify the total width and the shift amount. E.g. a `fixed.Shape(16, 12)` would have 4 integer bits and 12 fractional bits. Going this route, we could also consider dropping the `signed` argument and specifying e.g. `fixed.Shape(signed(16), 12)`, unambiguously specifying the shape of the underlying storage. `fixed.Shape(16, 20)` would also be more readable than `fixed.Shape(-4, 20)`.

jrmoserbaltimore:
> `fixed.Shape(16, 20)` would also be more readable than `fixed.Shape(-4, 20)`.

Check my math on the fixedShape ones.  I went over them a few times and got a few wrong and, honestly, I spend several seconds thinking about it but I'm not rigorously confirming they're right, it's effort.

```
              SQ        UQ        fixed.Shape (signed)  fixed.Shape (unsigned)
101.0011      SQ(4,4)   UQ(3,4)   fixed.Shape(8,4)      fixed.shape(7,4)
0.0001010011  SQ(-3,8)  UQ(-3,7)  fixedShape(8,8)       fixed.shape(7,8)
1010011000    SQ(8,-3)  UQ(7,-3)  fixed.Shape(8,-3)     fixed.Shape(7,-3)
```

Making this table, three things are obvious:

1. Coming up with SQ and UQ takes approximately no thought.  By the time I'm done typing the flat number, I've got all the information I need for SQ and UQ.
2. the fixed.Shape() notation takes significant mental effort *and* I got it wrong twice for the second example.
3. The third example has the same numbers in both formats.

It seems like the SQ/UQ format is far more readable and also far more writable.  Going both directions with the `(bits, shift)` notation requires mental math.  It actually reminds me of scientific notation; I stopped using scientific notation because I'd get exam questions wrong and burn 8 or 9 attempts on a homework question due to not properly shifting the decimal point around when doing arithmetic on dissimilar terms.

jrmoserbaltimore:
> * +1 that `fixed.Shape(7, 0, signed=True)` counterintuitively has the same width as `signed(8)`. Would be nice to find another solution

I'm a fan of knowing what size you're working with.  In programming languages, we don't get an extra bit for a signed 64-bit int versus an unsigned—we actually have things like int32 vs uint32 or whatever, depending on languages—I don't see why we need to hide the details here.  It seems even worse to hide those details here since we're specifying hardware.

>   * What should `fixed.Value() >> fixed.Value()` mean? Shift by the rounded integral component or throw an exception?

Throw an exception.  Shifting by a non-integer is absurd and attempting to handle this would just enable programmers being clever by using non-obvious design details of the language.  You can always set up a signal consisting of some set of bits in the fixed-width signal.

schnommus:
> Going this route, we could also consider dropping the signed argument and specifying e.g. fixed.Shape(signed(16), 12), unambiguously specifying the shape of the underlying storage. fixed.Shape(16, 20) would also be more readable than fixed.Shape(-4, 20).

The `fixed.Shape(signed(16), 12)` syntax LGTM, I like that the equivalent Amaranth shape of the underlying storage is the first argument of the signature.

> It seems like the SQ/UQ format is far more readable and also far more writable. 

As in the existing draft RFC, I think there is room for both forms:
- `UQ` and `SQ` remain as aliases to construct a `fixed.Shape`.
- The arguments of `SQ`, `UQ` are transformed into the equivalent (correct) `fixed.Shape` arguments on construction.

As long as we don't have to jump through hoops, I think we should try to expose the industry standard Q notation.

> I'm a fan of knowing what size you're working with. In programming languages, we don't get an extra bit for a signed 64-bit int versus an unsigned—we actually have things like int32 vs uint32 or whatever, depending on languages—I don't see why we need to hide the details here. It seems even worse to hide those details here since we're specifying hardware.

Agree, I think @zyp's suggestion above is a neat solution to this.

> Throw an exception. Shifting by a non-integer is absurd and attempting to handle this would just enable programmers being clever by using non-obvious design details of the language. You can always set up a signal consisting of some set of bits in the fixed-width signal.

Agree, to me it is obvious this operation should not be permitted.

text/0041-fixed-point.md
schnommus: @samimia-swks would you happen to have some references we can cite to show that this is the dominant Q notation in the audio ASIC world? It will be useful evidence that may be worth appending to the RFC document itself.

text/0041-fixed-point.md
schnommus: I would lean toward `lib.fixed`

whitequark: `lib.fixed` seems good, looking through the usage and discussion; `fixed.Shape`, `fixed.Value` are pretty natural to write.

vk2seb:
Based on all the helpful feedback above, I have an updated draft RFC -

- Draft: [Rendered](https://github.com/vk2seb/rfcs/blob/seb/fixed-point/text/0041-fixed-point.md)
- Individual changes: [Changelist](https://github.com/vk2seb/rfcs/commits/seb/fixed-point/text/0041-fixed-point.md)

@whitequark - would you prefer we try to stick to this PR (to preserve comment history), or would you prefer I open a new PR and we continue discussion there? I'm fine with either, although for the former I'd need the permissions.

FYI @zyp

samimia-swks:
This is looking good!

One note on the rounding topic:

Simple rounding is also known as a mid-tread quantizer. If the desired output format is I.Q (total length: I+Q) we add 0.5 / 2^(Q) and truncate. The cost is a real adder and a clipper if you wanna be safe. Not trivial if you didn’t care about the DC offset error that simple truncation introduces.
But you could also do a mid rise quantizer. You truncate to I.Q, but to get rid of the negative DC offset you add that 0.5 / 2^(Q) after truncation. This is technically an adder too but in HW you can just append an LSB of 1 which is free!
But now you end up with I+Q+1 bits. If these bits were going across a bus you would transport just the I.Q but the receiver has to remember to pad the LSB.

(Note that in integer land, mid-tread is like rounding to nearest integer and mid-rise is like rounding to the nearest midpoint between integers, so you cannot represent a hard zero)

With all that said, if you want to resize a number to I+Q bits without adding a DC offset AND you don’t want the cost of an adder, you can simply truncate to I.(Q-1) bits and then pad an LSB of 1!

So I would advocate for 3 options with truncate being default:

1. truncate (floors to -inf, free, adds dc offset)
2. mid-tread (rounds, costs an adder, can represent zero)
3. mid-rise (rounds, free, cannot represent zero)

On Nov 23, 2024, at 9:57 AM, Seb Holzapfel ***@***.***> wrote:

﻿
CAUTION: The email below is from an external source. Use caution before opening attachments or clicking links.



Based on all the helpful feedback above, I have an updated draft RFC -

  *   Draft: Rendered<https://urldefense.com/v3/__https://github.com/vk2seb/rfcs/blob/seb/fixed-point/text/0041-fixed-point.md__;!!MyQQGECaxY11k7S_!Zq4LoAYGXEt4HM1qPKfyQFLUr4EauoBPfcxvNS4aF5pj35yS3HS0dTRwo1OMRszu0B4aAl_mtwUL0cWq9SXnKMi1eHut8zw$>
  *   Individual changes: Changelist<https://urldefense.com/v3/__https://github.com/vk2seb/rfcs/commits/seb/fixed-point/text/0041-fixed-point.md__;!!MyQQGECaxY11k7S_!Zq4LoAYGXEt4HM1qPKfyQFLUr4EauoBPfcxvNS4aF5pj35yS3HS0dTRwo1OMRszu0B4aAl_mtwUL0cWq9SXnKMi10qhl9yU$>

@whitequark<https://urldefense.com/v3/__https://github.com/whitequark__;!!MyQQGECaxY11k7S_!Zq4LoAYGXEt4HM1qPKfyQFLUr4EauoBPfcxvNS4aF5pj35yS3HS0dTRwo1OMRszu0B4aAl_mtwUL0cWq9SXnKMi19f9ATEs$> - would you prefer we try to stick to this PR (to preserve comment history), or would you prefer I open a new PR and we continue discussion there? I'm fine with either, although for the former I'd need the permissions.

FYI @zyp<https://urldefense.com/v3/__https://github.com/zyp__;!!MyQQGECaxY11k7S_!Zq4LoAYGXEt4HM1qPKfyQFLUr4EauoBPfcxvNS4aF5pj35yS3HS0dTRwo1OMRszu0B4aAl_mtwUL0cWq9SXnKMi1ZTczUjI$>

text/0041-fixed-point.md
```diff
@@ -0,0 +1,177 @@
...
+- `.i_bits`, `.f_bits`, `.signed`: Width and signedness properties.
+- `.shape()`: Return the `fixed.Shape` this was created from.
+- `.as_value()`: Return the underlying value.
+- `.numerator()`: Return `as_value()` cast to the appropriate signedness.
+- `.eq(value)`: Assign `value`.
+  - If `value` is a `Value`, it'll be assigned directly to the underlying `Value`.
+  - If `value` is an `int` or `float`, it'll be cast to a `fixed.Const` first.
+  - If `value` is a `fixed.Value`, the precision will be extended or truncated as required.
+- `.reshape(f_bits)`: Return a new `fixed.Value` with `f_bits` fractional bits, truncating or extending precision as required.
+- `.reshape(shape)`: Return a new `fixed.Value` with shape `shape`, truncating or extending precision as required.
```
vk2seb, line 63: The 2 signatures of `.reshape()` here are my error. They should be collapsed into one `.reshape()`, although I am also considering dropping the second form if I can't find a killer use-case for it.

vk2seb:
Thank you samimia-swks@ for your input on the rounding details. This is all useful context, although at this point I am inclined to postpone rounding strategies for a future RFC.

In the latest version of the RFC, I have de-scoped rounding completely. 
Excerpt from the RFC text touching on this:
```
 - samimia-swks@: In most DSP applications, simple truncating is done (bit picking, which is equivalent to a floor()) because it's free. I would vote for that being the default behavior at least.
  - ld-cd@: (...) Truncate is still a reasonable default for most applications.
  - ld-cd@: (...) I think a better approach would be to leave rounding and several other common operations that require platform dependent lowering to a subsequent RFC (...).
  - vk2seb@: Both truncation and simple rounding (round to nearest) are commonly used in DSP algorithms. For now, we provide only `reshape()` (truncation, now reflected above). Additional rounding strategies may be added in a future RFC, however we will always need a default rounding strategy, and truncation seems like a sane default.
```

Even without an opinionated `round()` operation, I expect this RFC to be useful. Based on the above comments and experience from my own projects, truncation seems the most common 'form' of rounding (small sample size, so input here is appreciated!). Additionally, there is nothing preventing use of your own rounding strategies on top of the existing RFC, before they become part of the language in a future RFC.

text/0041-fixed-point.md
vk2seb: In any case I think this method could use further bikeshedding. I would be just as open to calling this method `truncate(f_bits=0)` or `trunc(f_bits=0)`, although that loses the notion that it can both increase *and* decrease precision, which was part of the reason of not calling things `round()` in the first place.

whitequark:
> In the latest version of the RFC, I have de-scoped rounding completely.

(I haven't been able to track the changes in depth but the overall approach sounds good. We have previously de-scoped controversial or poorly understood aspects of RFC and it worked quite fine.)

mndza:
Really happy to see progress on this RFC. I appreciate all the work that has gone into it.

I'm particularly interested in supporting user-selectable rounding modes beyond simple truncation. As noted by @vk2seb, truncation can introduce DC biases in the signal that are problematic for some DSP applications. However, I'm okay with that being left out of this particular RFC.

Besides that, I like the current proposal as it stands.

vk2seb:
Just a small update, I'm still committed to getting this RFC finished, just pushed up a functional implementation of the latest version of this RFC (with some test cases) here: https://github.com/amaranth-lang/amaranth/pull/1578

I'm currently experimenting with this in my own projects to make sure it interacts with the rest of Amaranth without any unexpected surprises. I aim to integrate any findings from that effort into this RFC and move it from draft to review state after fleshing out the missing sections. I can't commit to a hard deadline but would like to have this formally 'ready for review' within a month or so.

goekce:
I tried @vk2seb's implementation and noticed that `len(Value)` and `Shape.width()` are not part of the API. These are described as methods in the guide for standard shapes https://amaranth-lang.org/docs/amaranth/latest/guide.html#shapes . Don't you think it would be beneficial to include them in all shapes, including `fixed`?

```py
from amaranth import *
len(Const(2, unsigned(2)))  # 2
signed(4).width  # 4
```

vk2seb:
> I tried @vk2seb's implementation and noticed that `len(Value)` and `Shape.width()` are not part of the API. These are described as methods in the guide for standard shapes https://amaranth-lang.org/docs/amaranth/latest/guide.html#shapes . Don't you think it would be beneficial to include them in all shapes as a general API to get the number of bits in general, including `fixed`?
> 
> ```py
> from amaranth import *
> len(Const(2, unsigned(2)))  # 2
> signed(4).width  # 4
> ```

@goekce thanks, I agree that `fixed.Shape` should remain as similar to `Shape` in this regard and that these methods should also exist on `fixed.Shape`. will add this.

goekce:
Another observation:

If signal values are displayed, then they don't directly show a human-friendly representation like a `signed` signal:
```
In [486]: Const(-1)
Out[486]: (const 1'sd-1)

In [485]: fixed.Const(1.2, fixed.SQ(2,3))
Out[485]: fixed.SQ(2, 3) (const 5'sd10)
```
As I understand, `as_float()` must be used for this purpose:

```
In [487]: fixed.Const(1.2, fixed.SQ(2,3)).as_float()
Out[487]: 1.25
```

If it were to include 1.25 directly, then the representation would be more readable. However, all values build on either signed or unsigned values and `fixed` is implemented by applying a `fixed` shape lens to signed and unsigned values, as seen in the representation `fixed.SQ(2, 3) (const 5'sd10)`. Do I understand correctly?

vk2seb:
@goekce the precise implementation of `__repr__` is not described in the RFC text, I guess your comment is referring to the WIP implementation from https://github.com/amaranth-lang/amaranth/pull/1578

We can discuss what the output of `__repr__` should be for a `fixed.Value`, and add it to the RFC text. I don't necessarily think `fixed.SQ(2, 3) (const 5'sd10)` is perfect, and am open to other ideas. I do think it should be obvious at a glance that `fixed` acts as a lens on an underlying `Value`.

goekce:
> @goekce the precise implementation of `__repr__` is not described in the RFC text, I guess your comment is referring to the WIP implementation from [amaranth-lang/amaranth#1578](https://github.com/amaranth-lang/amaranth/pull/1578)

You are right, thanks!

> We can discuss what the output of `__repr__` should be for a `fixed.Value`, and add it to the RFC text. I don't necessarily think `fixed.SQ(2, 3) (const 5'sd10)` is perfect, and am open to other ideas. I do think it should be obvious at a glance that `fixed` acts as a lens on an underlying `Value`.

So you mean that `fixed.SQ(2, 3) (const 5'sd10)` could be the minimum and human-readable output should be integrated, right? Then an idea could be:

```
fixed.SQ(2, 3) (const 5'sd10) [1.25]
```

whitequark:
> So you mean that `fixed.SQ(2, 3) (const 5'sd10)` could be the minimum and human-readable output should be integrated, right? Then an idea could be:

The contract of `__repr__` is that the value should be printed in a way that makes it possible to reconstruct it by inserting it into the Python REPL (with some unspecified set of imports already done). So I'm against adding `[1.25]` as a suffix.

If there is a syntax for writing a compile-time fixed point constant then that syntax should be used. (It's been a while since I read the RFC, and I don't quite recall the details...) Otherwise, something like `<fixed.SQ(2, 3) ...>` should be used.

vk2seb:
> If there is a syntax for writing a compile-time fixed point constant then that syntax should be used. (It's been a while since I read the RFC, and I don't quite recall the details...) Otherwise, something like <fixed.SQ(2, 3) ...> should be used.

Thanks, this is helpful. In that case we want different implementations of `__repl__` for `fixed.Value` and `fixed.Const`. Then we can keep the logical (float) value in first position. Perhaps along the lines of:

```
>>> from amaranth import *
>>> from amaranth.lib import fixed
>>> fixed.Const(-0.75)
fixed.Const(-0.75, SQ(1, 2))
>>> fixed.Const(-0.75) * 3
fixed.Value(SQ(3, 2), (* (const 3'sd-3) (const 2'd3)))
# Check reconstruction property
>>> from amaranth.lib.fixed import SQ, UQ
>>> fixed.Const(-0.75, SQ(1, 2))
fixed.Const(-0.75, SQ(1, 2))
```

vk2seb:
> I also noticed `__str__` should be included in the RFC. Using `__str__` from the parent class could lead to confusion. In the following fixed signal is printed as -4:

I don't think that's necessary, as `__str__` will fall back to `__repr__`. However, we definitely want `fixed.Shape` to implement `.format()` so that the `Print` statement in your example works as expected.

whitequark:
This repository has been migrated to Codeberg. Please continue the discussion over at https://codeberg.org/amaranth-lang/rfcs/pulls/41. I am aware that this will unfortunately cause some degree of disruption; please contact me privately or on IRC/Matrix channels to resolve any issues.

PR #42, opened 2024-01-30
Add an RFC for `Const` from shape-castable.

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/const-from-shape-castable/text/0042-const-from-shape-castable.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-02-05#1707153339-1707153740;) this RFC on the 2024-02-05 weekly meeting. The disposition was to **merge**, without changes.

@galibert suggested that `Shape` should be `ShapeCastable` and `Value` should be `ValueCastable` to improve consistency within the language.

whitequark:
Thanks @zyp!

PR #43, opened 2024-01-31
Rename `reset=` to `init=`

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/rename-reset/text/0000-rename-reset-to-init.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-02-12#1707758183-1707758386;) this RFC on the 2024-02-12 weekly meeting. The disposition was to **merge**, without changes.

The deprecation period should be the same as for `Record`, two releases, since the fix is more trivial but the scope of the change is wider.

PR #44, opened 2024-01-31
Syntax for initial state property assertions

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/with-initial/text/0000-initial.md)

### DIFF
--- a/text/0000-initial.md
+++ b/text/0000-initial.md
@@ -1,0 +1,113 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [amaranth-lang/rfcs#44](https://github.com/amaranth-lang/rfcs/pull/44)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# Syntax for initial state property assertions
+
+## Summary
+[summary]: #summary
+
+Add dedicated syntax for describing assertions or assumptions for the initial simulation state.
+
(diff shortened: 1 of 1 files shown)


text/0000-initial.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Prior art
+[prior-art]: #prior-art
+
+This feature heavily draws from the Verilog `initial` construct.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Should this be a `with m.X():` control flow style construct or a `with m.x +=` statement style construct?
```
wanda-phi, line 106: `with m.x +=` → `m.x +=`

whitequark: fixed, thanks

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-02-12#1707758924-1707760772;) this RFC on the 2024-02-12 weekly meeting. As a result of the discussion I am withdrawing this RFC (and closing it in my capacity of the RFC author).

The reason for this is that I have become convinced that this is not the correct approach, and that `Initial`, or `m.initial +=`, or `with m.Initial():` are not meanignful constructs in their own right but simply an artifact of immaturity of the Yosys formal verification flows. We should improve Yosys instead of entrenching Yosys-specific workarounds. Commercial formal tools do not adopt anything like `$initstate`; they let you use a normal reset. We should aim to provide a similar experience.

PR #46, opened 2024-02-03
Add RFC for `Shape.cast(range(1))` change.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/shape-range-1/text/0046-shape-range-1.md)

Implementation PR: amaranth-lang/amaranth#1077.

### DIFF
--- a/text/0046-shape-range-1.md
+++ b/text/0046-shape-range-1.md
@@ -1,0 +1,58 @@
+- Start Date: 2024-02-05
+- RFC PR: [amaranth-lang/rfcs#46](https://github.com/amaranth-lang/rfcs/pull/46)
+- Amaranth Issue: [amaranth-lang/amaranth#1081](https://github.com/amaranth-lang/amaranth/issues/1081)
+
+# Change `Shape.cast(range(1))` to `unsigned(0)`
+
+## Summary
+[summary]: #summary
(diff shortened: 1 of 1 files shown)


whitequark:
I propose deprecating `bits_for` in the same swoop. Do we have simple guidance for replacing it?

wanda-phi:
> I propose deprecating `bits_for` in the same swoop. Do we have simple guidance for replacing it?

Not really. `x.bit_length()` works for unsigned numbers, but when negative numbers are involved you basically have to replicate the `bits_for` implementation (except for perhaps changing the rule for 0).

whitequark:
Maybe `C(x).shape.width`?

wanda-phi:
> Maybe `C(x).shape.width`?

That works, though it has the same 0 special case.

whitequark:
Yeah, but then it clearly falls out of the existing language rules.

Not sure how you can get the behavior where it always adds the sign bit though...

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-02-05#1707152662-1707152774;) this RFC on the 2024-02-05 weekly meeting. The disposition was to **merge**, without changes.

PR #48, opened 2024-02-19
Add RFC for passing a platform to the simulator.

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/simulator-platform/text/0048-simulator-platform.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-02-26#1708967140-1708968210;) this RFC on the 2024-02-26 weekly meeting. The disposition was to **close**, for the following reasons:
- The proposed functionality is already available via `Fragment.get(elaboratable, platform)`.
- It is not clear that it is desirable to be able to fully override the platform used by the simulator, once the simulator does provide a platform in the future.

PR #50, opened 2024-03-01
Add an RFC for `Print` statement and string formatting.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/print/text/0050-print.md)

### DIFF
--- a/text/0050-print.md
+++ b/text/0050-print.md
@@ -1,0 +1,204 @@
+- Start Date: 2024-03-04
+- RFC PR: [amaranth-lang/rfcs#50](https://github.com/amaranth-lang/rfcs/pull/50)
+- Amaranth Issue: [amaranth-lang/amaranth#1186](https://github.com/amaranth-lang/amaranth/issues/1186)
+
+# `Print` statement and string formatting
+
+## Summary
+[summary]: #summary
+
+A `Print` statement is added, providing for simulation-time printing. A `Format` object, based on Python format strings, is added to support the `Print` statement. The existing `Assert`, `Assume`, and `Cover` statements are modified to make use of `Format` functionality as well.
+
+## Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-04#1709571664-1709572235;) this RFC on the 2024-03-04 weekly meeting. The disposition was to **merge**, with the following changes:
- Clarify that the first character of a string is stored in the least significant bits of a string (which is the inverse of what Verilog does).
- Define `__format__` on Value to prevent unintentional misuse of f-strings that will not result in the desired effect (it will insert the AST dump instead of the runtime value into the string).

lekcyjna123:
Hi,
in "Future possibilities" you have pointed to:
> - Some kind of output redirection or output hook could be implemented in pysim.
> - Likewise, a hook for assertion failure could be desirable.
> - Some way to print the current simulation time could be useful.

Maybe you will be interested in https://github.com/kuznia-rdzeni/coreblocks/pull/595 where we have recently added logging possibility with syntax similar to the python `logging` library. There is integration with pysim and cocotb simulators. The printing is handled in separate simulation process so we have a possibility to inject custom callable on having error print. There is also simulation time printed.

The code is not very strictly bound to the transactron framework. Only the `DependencyContext` is used (something like global variable), so I think that it can be possible to integrate our solution with Amaranth if you wish.

@tilk @xThaid please look on this RFC and RFC 58.

whitequark:
Note that `Print` is designed to be synthesizable so that it can also be used with CXXRTL or exported to Verilog for use with Verilator, etc. It is explicitly not desired to introduce features that would only be implementable in a Python simulation but not in these other contexts.

In light of this, I think your logging functionality would probably need to reserve the beginning of the string for metadata, which can be then filtered by means of capturing and processing the output.

PR #51, opened 2024-03-03
Add an RFC for `ShapeCastable.from_bits` and `lib.data.Const`.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/literal/text/0051-from-bits-literalview.md)

### DIFF
--- a/text/0051-from-bits-literalview.md
+++ b/text/0051-from-bits-literalview.md
@@ -1,0 +1,152 @@
+- Start Date: (fill me in with today's date, 2024-03-04)
+- RFC PR: [amaranth-lang/rfcs#51](https://github.com/amaranth-lang/rfcs/pull/51)
+- Amaranth Issue: [amaranth-lang/amaranth#1187](https://github.com/amaranth-lang/amaranth/issues/1187)
+
+# Add `ShapeCastable.from_bits` and `amaranth.lib.data.Const`
+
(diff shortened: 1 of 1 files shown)


whitequark:
Would `from_bits` ever need to know the size of the original value?

wanda-phi:
> Would `from_bits` ever need to know the size of the original value?

It is called on `ShapeCastable`, which has the length available (since it has to implement `as_shape()`).

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-04#1709572254-1709574200;) this RFC on the 2024-03-04 weekly meeting. The disposition was to **merge**, with the following changes:
- `lib.data.LiteralView` is renamed to `lib.data.Const` and made immutable;
- `ShapeCastable.from_bits` is allowed to return any type as long as `ShapeCastable.const` can turn it back into a const-castable value;
  - Additional unresolved question: With the *same* bit pattern or *equivalent* bit pattern? This is important if we ever try to roundtrip a NaN through Python's `float`.
- `lib.data.Layout.const` is changed to return a `lib.data.Const`;
- Mentioning that instead of mutating a `lib.data.Const` in simulation, one can set the field directly.

whitequark:
Thank you!

PR #52, opened 2024-03-05
Add RFC for `Choice`.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/choice/text/0052-choice.md)

### DIFF
--- a/text/0052-choice.md
+++ b/text/0052-choice.md
@@ -1,0 +1,183 @@
+- Start Date: 2024-07-01
+- RFC PR: [amaranth-lang/rfcs#52](https://github.com/amaranth-lang/rfcs/pull/52)
+- Amaranth Issue: [amaranth-lang/amaranth#1445](https://github.com/amaranth-lang/amaranth/issues/1445)
+
+# Add `amaranth.hdl.Choice`, a pattern-based `Value` multiplexer
+
+## Summary
+[summary]: #summary
+
+A new type of expression is added: `amaranth.hdl.Choice`. It is essentially a variant of `m.Switch`
+that returns a `Value` using the same patterns as `m.Case` for selection.
+
+## Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


whitequark:
Guard syntax bikeshed:

1. Just stuff it in the list:

```
Choice(foo, {
  (1, guard): bar
})
```

2. Context manager:

```
with Choice(foo) as c:
    with c.Case(1, if=guard):
        c.eq(bar)
```

(I don't love any of these. They're also both backward compatible.)

wanda-phi:
> Guard syntax bikeshed:
> 
>     1. Just stuff it in the list:
> 
> 
> ```
> Choice(foo, {
>   (1, guard): bar
> })
> ```

That won't work (`Value` is not hashable)

whitequark:
In addition, the existing Mux expression is made valid on the left-hand side of an assignment, as if it was lowered as follows:

```py
def Mux(sel, val1, val0):
    return Choice(a).case(1, val1).default(val0)
```

Shouldn't it be the opposite?

```py
def Mux(sel, val1, val0):
    return Choice(a).case(0, val0).default(val1)
```

wanda-phi:
> > In addition, the existing Mux expression is made valid on the left-hand side of an assignment, as if it was lowered as follows:
> > ```python
> > def Mux(sel, val1, val0):
> >     return Choice(a).case(1, val1).default(val0)
> > ```
> 
> Shouldn't it be the opposite?
> 
> ```python
> def Mux(sel, val1, val0):
>     return Choice(a).case(0, val0).default(val1)
> ```

fixed, thanks

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-07-01#1719856007-1719856763;) this RFC on the 2024-07-01 weekly meeting. The disposition was to **merge**.

whitequark:
Thanks @wanda-phi for your work!

PR #53, opened 2024-03-06
RFC 53: Low-level I/O primitives.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/ioport/text/0053-ioport.md)

### DIFF
--- a/text/0053-ioport.md
+++ b/text/0053-ioport.md
@@ -1,0 +1,221 @@
+- Start Date: 2024-03-11
+- RFC PR: [amaranth-lang/rfcs#53](https://github.com/amaranth-lang/rfcs/pull/53)
+- Amaranth Issue: [amaranth-lang/amaranth#1195](https://github.com/amaranth-lang/amaranth/issues/1195)
+
+# Low-level I/O primitives
+
+## Summary
+[summary]: #summary
+
+A new first-class concept is introduced to the language, I/O ports, representing top-level ports in synthesis flows. I/O ports can be obtained from the platform, or can be manually constructed by the user in raw Verilog flows that don't use a supported platform.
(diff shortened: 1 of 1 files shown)


whitequark:
> Is a deprecation period even necessary for Values as io ports? Arguably, all well-behaved code will continue to just work given that platform.request(dir='-') returns an IOPort.

After sleeping on it: we can probably do this right away. I don't think we have any uncursed way of obtaining a toplevel `inout` without a platform or the newly introduced undocumented `ports=` alternate API, do we?

> Should we forbid mixing various ways of port creation?

Eventually probably, but I feel like this can be left until a later point when we improve the backend interface as a whole.

> Traditional name bikeshedding.

I'm happy with `IOValue`/`IOPort`.

mwkmwkmwk:
Updated to make `metadata` per-bit.

wanda-phi:
Updated to make `metadata` per-bit.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-11#1710176822-1710177864;) this RFC on the 2024-03-11 Amaranth core meeting. The disposition was to **merge**.

text/0053-ioport.md
```diff
@@ -0,0 +1,221 @@
+- Start Date: 2024-03-11
+- RFC PR: [amaranth-lang/rfcs#53](https://github.com/amaranth-lang/rfcs/pull/53)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
```
whitequark, line 3: Needs a tracking issue.

PR #55, opened 2024-03-07
RFC 55: New `lib.io` components.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/lib-io/text/0055-lib-io.md)

### DIFF
--- a/text/0055-lib-io.md
+++ b/text/0055-lib-io.md
@@ -1,0 +1,171 @@
+- Start Date: 2024-03-18
+- RFC PR: [amaranth-lang/rfcs#55](https://github.com/amaranth-lang/rfcs/pull/55)
+- Amaranth Issue: [amaranth-lang/amaranth#1210](https://github.com/amaranth-lang/amaranth/issues/1210)
+
+# New `lib.io` components
+
+## Summary
+[summary]: #summary
+
+Building on RFC 2 and RFC 53, a new set of components is added to `lib.io`. The current contents of `lib.io` (`Pin` and its signature) become deprecated.
+
+## Motivation
+[motivation]: #motivation
+
+Currently, all IO buffer and register logic is instantiated in one place by `platform.request`. Per [amaranth-lang/amaranth#458](https://github.com/amaranth-lang/amaranth/issues/458), this has caused problems. Ideally, the act of requesting a specific I/O (user's responsibility) would be decoupled from instantiating the I/O buffer (peripherial library's responsibility).
+
+Further, we currently have no standard I/O buffer components, other than the low-level `IOBufferInstance`.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+`IOPort`, introduced in RFC 53, is Amaranth's low-level IO port primitive. In `lib.io`, `IOPort`s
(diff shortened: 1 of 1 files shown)


whitequark:
(I'll nominate this RFC for discussion if/when #53 is accepted.)

wanda-phi:
> Are we losing anything by removing `DiffPair` from the standard library entirely?

There are vendor cells (that a user may want to use with `Instance`) that want both pins of a diff pair. I think you can get away with just skipping the complement one in many cases, but I wouldn't want to rely on that working for all platforms we want to support.

whitequark:
> There are vendor cells (that a user may want to use with `Instance`) that want both pins of a diff pair. I think you can get away with just skipping the complement one in many cases, but I wouldn't want to rely on that working for all platforms we want to support.

I am aware. I think this should be the odd platform's problem rather than making everyone track the useless complement pin just to have it be discarded. There are platforms where even its assignment isn't verified at all (iCE40)--this is worse than useless!

wanda-phi:
> > There are vendor cells (that a user may want to use with `Instance`) that want both pins of a diff pair. I think you can get away with just skipping the complement one in many cases, but I wouldn't want to rely on that working for all platforms we want to support.
> 
> I am aware. I think this should be the odd platform's problem rather than making everyone track the useless complement pin just to have it be discarded. There are platforms where even its assignment isn't verified at all (iCE40)--this is worse than useless!

idk, I'm not super in love with `DiffPair`, but my feeling on that is that returning the `DiffPair` is cleaner even if the neg half of the pair ends up useless in practice

If truly no platform ends up needing the neg half connected, you're right and there's no point to it. However, if a platform does come up where you actually need to connect the neg port to some instance pin, it's a bit of a mess: the platform code handling the `*Buffer` (or even user code making a custom `Instance`) needs to introspect the `IOValue` (which may be a slice or concat), get all relevant `metadata`s, and reassemble them back to a concat/slice of the neg `IOPort`s. This would require exposing a bunch of internal AST details to any code that may need to do that. Meanwhile, with diffpairs we can just have slice/concat operations on the `DiffPair` object that make this work automatically.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-11#1710177904-1710180626;) the RC on the 2024-03-11 Amaranth core meeting. The disposition was to **postpone**.

whitequark:
After considering that not having this functionality is a direct blocker to any development of standard I/O cores (as the methodology requires it), I'm nominating this for discussion again on the next meeting, under the assumption that @wanda-phi adds buffer directionality checking to `*Port` classes.

whitequark:
Bikeshed: `dir` is a name that exists in Python toplevel (`dir()`, a debugging function). I propose that in all of the newly introduced APIs, where the use will be almost entirely positional (or limited to custom implementations of buffers, which aren't common), we call the direction `directoin`.

wanda-phi:
> Bikeshed: `dir` is a name that exists in Python toplevel (`dir()`, a debugging function). I propose that in all of the newly introduced APIs, where the use will be almost entirely positional (or limited to custom implementations of buffers, which aren't common), we call the direction `direction`.
> 
> To be clear, we leave `platform.request()` untouched by this.

Agreed, applied.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-18#1710782669-1710782985;) this RFC on the 2024-03-18 weekly meeting. The disposition was to **merge**.

PR #57, opened 2024-03-12
RFC 57: Single-Field Register Definition and Usage Shortcut

tpwrules:
[Rendered](https://github.com/tpwrules/amaranth-rfcs/blob/single-field-shortcut/text/0057-single-field-shortcut.md)

### DIFF
--- a/text/0057-single-field-shortcut.md
+++ b/text/0057-single-field-shortcut.md
@@ -1,0 +1,173 @@
+- Start Date: 2024-03-15
+- RFC PR: [amaranth-lang/rfcs#57](https://github.com/amaranth-lang/rfcs/pull/57)
+- Amaranth SoC Issue: [amaranth-lang/amaranth-soc#78](https://github.com/amaranth-lang/amaranth-soc/issues/78)
+
+# Single-Field Register Definition and Usage Shortcut
+
+## Summary
+[summary]: #summary
+
+Add a shortcut to `csr.Register` to enable easier and cleaner definition and use
+of registers which contain exactly one `csr.Field`.
+
+## Motivation
+[motivation]: #motivation
+
+Currently, use of a `csr.Register` in an amaranth-soc peripheral requires creation of a map (or list) of `csr.Field`s, which actually represent the register data and control access to each part. This is commonly done by creating a `Register` subclass. Each `Field` also needs a name (or index) in the `Register`. However, many registers in peripherals only contain one field, so the extra name and subclass are unnecessary and redundant. This RFC introduces a shortcut for constructing and using a `Register` containing exactly one un-named `Field`.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Consider the following simple peripheral which generates a toggling signal on bits of an output corresponding to bits set in a register:
+
+```python
+from amaranth import *
+from amaranth.lib import wiring
+from amaranth.lib.wiring import In, Out
+from amaranth_soc import csr
+
+class Toggle(wiring.Component):
+    bus: In(csr.Signature(addr_width=1, data_width=32))
+    toggle: Out(32)
+
+    class Toggle(csr.Register, access="rw"):
+        toggle: csr.Field(csr.action.RW, 32)
+
+    def __init__(self):
+        self._toggle = Toggle()
+
+        # ... bridge setup ...
+
+    def elaborate(self, platform):
+        m = Module()
+
+        # ... bridge setup ...
+
+        for i in range(32):
+            with m.If(self._toggle.f.toggle.data[i]):
+                m.d.sync += self.toggle[i].eq(~self.toggle[i])
+            with m.Else():
+                m.d.sync += self.toggle[i].eq(0)
+
+        return m
+```
+
+The `toggle` name is used, among other ways, to name the register class, to name
+the field within, and to access that field on the register instance during
+elaboration.
+
+This can be simplified by passing the `csr.Field` directly to the `csr.Register`
+without enclosing it in a dict/list and naming/indexing it, or subclassing
+`csr.Register`. This also causes the `.f` member to access that Field directly
+instead of needing to provide a name/index.
+
+This simplifies the design as follows, reducing clutter and redundancy:
+
+```python
+from amaranth import *
+from amaranth.lib import wiring
+from amaranth.lib.wiring import In, Out
+from amaranth_soc import csr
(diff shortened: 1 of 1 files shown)


text/0057-single-field-shortcut.md
```diff
@@ -0,0 +1,167 @@
...
+    * `access` is either a `csr.Element.Access` value, or `None`.
+- a `.fields` property, returning the instantiated `FieldActionMap`/`FieldActionArray`/`FieldAction`;
+- a `.f` property, as a shorthand to `self.fields`;
+- a `.__iter__(self)` method that yields, for each contained field, a tuple containing its path (as a tuple of names or indices) and its instantiated `FieldAction`. If only a single `Field` was supplied, its path is always `tuple()`.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+* A novice might not understand that it's possible to create a `Register` with multiple `Fields` if exposed only to code which uses single-`Field` `Register`s.
+* There is a difference between a `Register` with one named/indexed `Field` and with one unnamed `Field`.
```
whitequark, line 120: This seems fine since the former suggests the register may be expanded in the future but the latter suggests the register will stay exactly as it is.

text/0057-single-field-shortcut.md
```diff
@@ -0,0 +1,167 @@
...
+* Library code will be made more confusing by having `fields` refer to possibly only one field.
```
whitequark, line 121: This could be used by having `.f` and `.field` only.

text/0057-single-field-shortcut.md
```diff
@@ -0,0 +1,167 @@
...
+* Downstream code will have to deal with a third type of object returned by the
+  `.f` property which behaves unlike the current two.
```
whitequark, line 123: This is just a direct (desirable) consequence.

text/0057-single-field-shortcut.md
```diff
@@ -0,0 +1,167 @@
...
+* BSP/doc generators will have to intelligently render a `Field` with an empty name and path (likely by omitting them, possibly by reusing the register's name).
```
whitequark, line 124: This is just a direct (desirable) consequence.

text/0057-single-field-shortcut.md
```diff
@@ -0,0 +1,167 @@
...
+    * An empty name is prohibited by `FieldActionMap`, so empty names are a novel
+      problem.
+    * The register does not know its own name independently of its container, so this may be difficult in practice.
```
whitequark, line 127: No this is not difficult, this is actually how this should be handled to get an actually good BSP generated. This isn't at all a downside, but rather an improvement on the status quo that we can now do this.

text/0057-single-field-shortcut.md
```diff
@@ -0,0 +1,167 @@
...
+AVR microcontrollers have many registers where the field name is the same as the register name, though this is not always the case for single-field registers. Application code usually uses the register name in this case and ignores the field name.
+
+STM32 microcontrollers are similar, but the field name is usually a suffix of the register name, as the field name is not expected to be globally unique. However, the generated BSP files do appear to always contain the field name at least somewhere.
+
+More experienced input is welcomed here.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* Should we change `fields` to `field` in the constructor and properties? This could break existing code.
```
whitequark, line 159: Basically everyone uses `.f`, but yeah.

text/0057-single-field-shortcut.md
```diff
@@ -0,0 +1,167 @@
...
+* Should we add `field` too? This would access the same object/s as `fields` and so would make available the option to use the right plurality, though it would be more code and wouldn't force it.
+* Should we force a `Register` with a single `Field` to always have that one be un-named? Would create backwards compatibility problems, but reduce ambiguity.
```
whitequark, line 161: This would be a regression in the ability to make backwards compatible changes.

text/0057-single-field-shortcut.md
```diff
@@ -0,0 +1,167 @@
...
+* Should we go further? `Register` could take a single `FieldAction` and automatically wrap it in a `Field` (and inherit its access mode).
```
whitequark, line 162: This seems like it would make the DSL unnecessarily confusing.

text/0057-single-field-shortcut.md
tpwrules: I think it's more likely to be a problem with the constructor argument.

text/0057-single-field-shortcut.md
tpwrules: Yes, it would make the constructor arguments excessively clever, but it could be done.

The benefit would be not needing Field in many cases, but that could also further entrench single-field registers.

whitequark: From what I've seen in the wild, people *way* overuse single-field registers, so I don't think encouraging that is a good idea.

text/0057-single-field-shortcut.md
jfng: Breakage isn't an issue yet (until 0.1 is released).

I see three possibilities:
a) do nothing;
b) add a `.field` property, aliased to `.fields`;
c) rename `.fields` to `.field`, but keep the `fields=` parameter;
d) do c) and rename `fields=` to `field=`;

Option c) has my preference.

Considering the example given in this RFC, a single-field register won't pass `fields=` as a keyword argument (`csr.Register(csr.Field(...), access="rw")`), so having it plural is mostly invisible.

OTOH the plural `.fields` accessor is awkward for a single-field register.

jfng:
This RFC was discussed during the [2024-03-15](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-15#1710522067-1710523924;) SoC meeting.

The decision is **merge**, with the following changes:
- rename `Register.fields` to `.field`.

whitequark:
@jfng Ping me once the RFC is updated to address the conversation (there should be no "unresolved questions" left) and I'll merge.

text/0057-single-field-shortcut.md
tpwrules: We settled that the property will be `.field` and the constructor will be `fields=`.

text/0057-single-field-shortcut.md
tpwrules: We settled that the property will be `.field` and the constructor will be `fields=`.

PR #58, opened 2024-03-13
Add an RFC for `ValueCastable` formatting.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/valuecastable-format/text/0058-valuecastable-format.md)

### DIFF
--- a/text/0058-valuecastable-format.md
+++ b/text/0058-valuecastable-format.md
@@ -1,0 +1,123 @@
+- Start Date: 2024-03-25
+- RFC PR: [amaranth-lang/rfcs#58](https://github.com/amaranth-lang/rfcs/pull/58)
+- Amaranth Issue: [amaranth-lang/amaranth#1243](https://github.com/amaranth-lang/amaranth/issues/1243)
+
+# Core support for `ValueCastable` formatting
+
+## Summary
+[summary]: #summary
+
+`Format` hooks are added, allowing custom formatting to be implemented for `ValueCastable`s.
+
+This RFC is only about adding hook support for custom `ShapeCastable`s. Providing actual formatting implementation for `lib.data` and `lib.enum` is left for a future RFC.
+
+## Motivation
+[motivation]: #motivation
+
+Custom types, like enums and data structures, will not make immediate sense to the user when printed as raw bit patterns. However, it is often desirable to print them out with the `Print` statement. Thus, an extension point for custom formatting needs to be added to the `Format` machinery.
+
+## Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+`ShapeCastable` subtypes can define their own ways of formatting values by implementing the `format` hook:
+
+```py
+class FixedPoint(ShapeCastable):
+    ...
+
+    def format(self, value, format_desc):
+        if format_desc == "b":
+            # Format a fixed-point value as binary.
+            return Format("{value.int:b}.{value.fract:0{fract_bits}b}}", value=value, fract_bits=len(value.fract))
+        elif format_desc == "x":
+            # Format a fixed-point value as hexadecimal (very simplified implementation).
+            assert len(value.fract) % 4 == 0
+            return Format("{value.int:x}.{value.fract:0{fract_digits}x}}", value=value, fract_digits=len(value.fract) // 4)
+        else:
+            ...
+
+# A fixed-point number with 8 integer and 8 fractional bits.
+num = Signal(FixedPoint(8, 8))
+m.d.comb += [
+    num.eq(0x1234)
+    Print(Format("Value in binary: {:b}", num)),
+    Print(Format("Value in hexadecimal: {:x}", num)),
+]
+```
+
+prints:
+
+```
+Value in binary: 00010010.00110100
+Value in hexadecimal: 12.34
+```
+
+However, sometimes it is also useful to print the raw value of such a signal. A shorthand is provided for that purpose:
+
+```py
+m.d.comb += Print(Format("Value: {num:x} (raw: {num!v:x})", num=num))
+```
+
+```
+Value: 12.34 (raw: 1234)
+```
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A new overridable method is added to `ShapeCastable`:
+
+- `format(self, value: ValueCastable, format_desc: str) -> Format`
+
+When a `ValueCastable` is formatted without specifying a conversion (ie. `"!r"` or `"!v"`):
+
+- `shape()` is called
+- if the shape is a `ShapeCastable` and has a `format` method, the value-castable being formatted and the format descriptor (the part after `:` in `{}`, if any) are passed directly to `shape.format()`, and the result (which must be a `Format` object) is included directly in place of the format specifier
+- otherwise (the shape is a plain `Shape`, or doesn't have `format`), `Value.cast()` is called on the value-castable, and formatting proceeds as if it was a plain value
+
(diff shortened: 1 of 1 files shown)


text/0058-valuecastable-format.md
```diff
@@ -0,0 +1,93 @@
...
+    Print(Format("current opcode: {}", op)),
+]
+```
+
+It is desirable that this example would print something like `"XOR"`, instead of `"2"`. Such functionality can be added on `ShapeCastable` subclasses by overriding the `format` method.
+
+However, sometimes it is also useful to print the raw value of such a signal. A shorthand is provided for that purpose:
+
+```py
+m.d.comb += Print(Format("current opcode: {op} (raw: {op!v})", op=op))
```
whitequark, line 42: Bikeshed: I feel weird about `!v`. I guess it's `v` for `Value.cast`, but I feel that `!i` for "turn it into an Integer" or `!n` for "turn it into a Number` might look less out of place.

text/0058-valuecastable-format.md
```diff
@@ -0,0 +1,93 @@
...
+The default implementation of `format` raises a `TypeError` exception.
+
+A new conversion, `!v`, is added to `Format` (with syntax analogous to `!r`). When specified, the value being formatted is converted through `Value.cast` before proceeding with further formatting. It can be used to print the raw value of value-castables. It is a no-op when used with a plain `Value`.
+
+An implementation of `__format__` is added to the `ValueCastable` base class that always raises a `TypeError`, directing the user to `Format` instead (like the one that already exists on plain `Value`).
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+None.
```
whitequark, line 67: The fact that a new name `format` is reserved on `ShapeCastable` and a default implementation for `__format__` is added to `ValueCastable` are definitely drawbacks as we are trying to keep those interfaces slim.

I feel especially conflicted about `__format__` on `ValueCastable`. That will be the first method with an actual implementation added on that interface. Is this the right thing to do?

text/0058-valuecastable-format.md
```diff
@@ -0,0 +1,93 @@
...
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+What should be the default implementation of the `format` hook?
+
+The alternatives are:
+
+- (proposed above) raise `TypeError`
+- defer to plain value formatting
+- do not provide one, require implementing the hook
```
whitequark, line 88: I don't think `format` is worth another breaking change. (`shape` was already a judgement call.)

text/0058-valuecastable-format.md
whitequark: `!r` would be perfect.

text/0058-valuecastable-format.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+What should be the default implementation of the `format` hook?
+
+The alternatives are:
+
+- (proposed above) raise `TypeError`
+- defer to plain value formatting
```
whitequark, line 87: I think this is probably the way to go. My reasoning is: imagine you are writing generic code, a FIFO or something. You want to display an element that is being popped. Your options:
- do not use `!v` and force consumers to sometimes implement a no-op `format` just so they can use your FIFO with debugging enabled;
- use `!v` and deny consumers which implement a special `format` legibility;
- branch on the exception or something (cursed, and probably what will actually take place eventually)

whitequark: This option is also the one which meshes best with our treatment of `ValueCastable`s otherwise (where they just decay to `Value`s unless something interferes with it).

wanda-phi: This is very convincing, I'll update the RFC.

whitequark:
Also, I think `lib.data` and `lib.enum` are kind of weird as motivation goes, since they should not be using this facility either at all or at least this facility alone, as they need to affect the generated Verilog (by enriching it with additional wires or attributes respectively).

text/0058-valuecastable-format.md
whitequark: ... `!R`?

whitequark: or `!av` for `.as_value()`

whitequark: or `!l` for "lower to a value"

wanda-phi:
> Also, I think `lib.data` and `lib.enum` are kind of weird as motivation goes, since they should not be using this facility either at all or at least this facility alone, as they need to affect the generated Verilog (by enriching it with additional wires or attributes respectively).

I feel these are orthogonal concerns. Regardless of what happens in Verilog, you still want *something* to happen when these are printed in simulation, perhaps even something further overridable by the user on a `Struct` subclass.

However, if you wish to postpone this RFC until we come up with holistic design for `lib.data` / `lib.enum` formatting and `_repr` replacement, I'm very much fine with this decision.

wanda-phi:
> > However, if you wish to postpone this RFC until we come up with holistic design for `lib.data` / `lib.enum` formatting and `_repr` replacement, I'm very much fine with this decision.
> 
> I think I'm happy to proceed for the reason you stated, I just feel like the motivation section should put more emphasis on non-magical user-defined data types like `lib.fixed`.
> 
> Also you make a good point re: subclasses, formatting should definitely be overridable for `Struct` and _especially_ `Union` (where the default one doesn't make a huge amount of sense).

The problem with `lib.fixed` formatting specifically is that it is not reasonably implementable in the way you'd want (eg. as a decimal number, or supporting all the weird format description stuff)

... I guess I can do some simple binary-only printout?

text/0058-valuecastable-format.md
wanda-phi: `.as_value()` is distinct from `Value.cast` (it can return another `ValueCastable`)

... actually this may be legitimate alternate semantics

whitequark:
> The problem with `lib.fixed` formatting specifically is that it is not reasonably implementable in the way you'd want (eg. as a decimal number, or supporting all the weird format description stuff)

I suppose not, but we want *some* formatting for fixpoint numbers that's better than just "print the number of LSBs in decimal" because that's atrocious.

text/0058-valuecastable-format.md
```diff
@@ -0,0 +1,128 @@
...
+The `format` hook is added on `ShapeCastable` instead of `ValueCastable`. This ensures that `ValueCastable`s without a custom shape automatically get plain formatting.
+
+The default behavior proposed in this RFC ensures that a formatting implementation is always available, allowing generic code to print arbitrary values without worrying about an exception. Eg. something like `lib.fifo` could use debug prints when an element is added, automatically using rich formatting for shapes with `format`, while falling back to plain integers when rich formatting is not available.
+
+alternative default behaviors possible are:
+
+- raise `TypeError` (disallow formatting value-castables without explicitly implemented formatting)
+- no default, require every `ShapeCastable` to implement `format` (compatibility breaking)
+
+To avoid reserving a name and interfering with user code, `format` could be renamed to `_amaranth_format`.
```
whitequark, line 101: If it ever becomes a problem (no idea if it will) I think we should add something that works on all of the special methods; either "a `amaranth_` prefixed version takes priority" or "if `_amaranth_namespace_ = True` is defined on the class look for different methods".

However, I think it's not the place of this RFC to introduce this namespacing.

wanda-phi: Agreed.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-25#1711388992-1711390083;) this RFC on the 2024-03-25 weekly meeting. The disposition was to **merge**.

We have not reached a consensus for the format character to be used for the `!v` syntax, but it should be using `Value.cast()` for conversion. The syntax will be further bikeshedded in the tracking issue, defaulting to `!v` if no consensus emerges there still.

PR #59, opened 2024-03-17
Add an RFC for removing upwards propagation of clock domains

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/domain-prop/text/0059-no-domain-upwards-propagation.md)

### DIFF
--- a/text/0059-no-domain-upwards-propagation.md
+++ b/text/0059-no-domain-upwards-propagation.md
@@ -1,0 +1,66 @@
+- Start Date: 2024-03-25
+- RFC PR: [amaranth-lang/rfcs#59](https://github.com/amaranth-lang/rfcs/pull/59)
+- Amaranth Issue: [amaranth-lang/amaranth#1242](https://github.com/amaranth-lang/amaranth/issues/1242)
+
+# Get rid of upwards propagation of clock domains
+
+## Summary
+[summary]: #summary
+
+Upwards propagation of clock domains is removed, `local=True` effectively becomes the default.
+
+## Motivation
+[motivation]: #motivation
+
+Currently, a clock domain created anywhere in the hierarchy is, by default, propagated everywhere within that hierarchy. The other option is marking a clock domain with `local=True`, which restricts propagation to only downwards from the point of definition.
+
+The default behavior constitutes the ultimate implicit spooky action at a distance, making it hard to reason about clock domain origins in complex Amaranth design. It is also rarely used. This dangerous default should be changed to something more reasonable.
+
+## Guide and reference-level explanation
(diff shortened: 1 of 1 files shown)


text/0059-no-domain-upwards-propagation.md
```diff
@@ -0,0 +1,73 @@
...
+## Guide and reference-level explanation
+[explanation]: #explanation
+
+In version 0.5, upwards domain propagation becomes deprecated. Any use of a domain that wouldn't be valid without upwards propagation (ie. wouldn't be valid if the domain were `local=True`) triggers a deprecation warning.
+
+In version 0.6, upwards domain propagation is removed entirely. The `local` argument to `ClockDomain` constructor becomes redundant and is deprecated.
+
+In version 0.7, the `local` argument is removed.
+
+For the cases where upwards domain propagation was used intentionally, a new mechanism is introduced to explicitly import a clock domain from a submodule:
```
whitequark, line 28: I think this is completely unnecessary. We should not be adding more `getattr`-like functionality where we can avoid it.

Is there a value-add over `m.domains.sync = clkgen.cd_sync`?

wanda-phi: Slight but non-zero: with this proposal, `clkgen` can defer creating `cd_sync` until `elaborate`, where it can decide things (including polarity) based on the platform

whitequark: Right. I would expect clock/reset generators to be so platform-specific in first place that the clock/reset generator itself would be created in `elaborate`, so this advantage is fairly slim.

whitequark:
More generally, since we have a pending full redesign of the system, we should probably not add functionality we might have to deprecate later, if it's avoidable.

text/0059-no-domain-upwards-propagation.md
```diff
@@ -0,0 +1,66 @@
...
+[unresolved-questions]: #unresolved-questions
+
+None.
+
+## Future possibilities
+[future-possibilities]: #future-possibilities
+
+The concept of "private" clock domains has been proposed: clock domains that don't propagate at all, including downwards.
+
+It's clear that clock domain overhaul with a larger scope needs to be done.
```
whitequark, line 66: Worth mentioning that this RFC is going to enable such overhaul (which may just not be feasible without it at all).

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-25#1711388626-1711388965;) this RFC on the 2024-03-25 weekly meeting. The disposition was to **merge**.

PR #61, opened 2024-03-20
Minimal stream interface

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/minimal-streams/text/0061-minimal-streams.md)


text/0061-minimal-streams.md
```diff
@@ -0,0 +1,266 @@
...
+- `amaranth.lib.stream.Signature(payload_shape, *, always_valid=False, always_ready=False)`
+    - with members:
+        - `payload: Out(payload_shape)`
+        - `valid: Out(1)`
+        - `ready: In(1)`
+    - with properties:
+        - `always_valid: bool`
+        - `always_ready: bool`
+    - with methods:
+        - `__eq__(other)`: requires `isinstance(other, Signature)` and all three of `payload_shape`, `always_valid`, `always_ready` to match
```
wanda-phi, line 59: please clarify that it's `lib.stream.Signature` in particular

whitequark: done

text/0061-minimal-streams.md
```diff
@@ -0,0 +1,266 @@
...
+7. Avalon-ST includes an unusual `error` signal that allows the source signal an error to the sink. Importantly, it is not useful for signaling protocol errors (e.g. overflow due to an error in `ready` signal handling by the source), since the sink either drives only a `ready` signal or nothing.
+
+The protocol in this proposal and the Avalon-ST protocol are interoperable if the Avalon-ST interface is configured with parameters `readyLatency = readyAllowance = 0` and includes `data`, `ready`, and `valid` signals only (i.e. beats, packets, or channels are not used). In addition, `data` must be renamed to `payload` and a posedge clock domain must be used in Amaranth.
+
+The Avalon-ST protocol is widely used in the industry, especially in the Altera (Intel (Altera)) ecosystem, and the ability to define components interoperable with it is highly desirable. While it could be used as-is in the Amaranth ecosystem, it is valuable to reconsider whether the inclusion of some of the rarely used features is worth the support and compatibility burden, since all generic interconnect would have to be aware of them. Splitting of `data` into symbols or bytes seems particularly unnecessary.
```
wanda-phi, line 193: :blue_heart:

daniestevez:
I've read through the RFC and it looks quite good to me. It captures well the simplest subset of AXI-Stream that is commonly used to move data around.

I'm curious about why asserting valid combinatorially is allowed in the following:

> The receiver may assert `ready` in response to the transmitter asserting `valid` (including combinatorially).

Specially in combination with

> The rules for asserting TREADY and TVALID in AXI-Stream are the same as the rules for asserting `ready` and `valid` in this proposal, except for the deassertion of valid during reset.

The [AXI specification](https://developer.arm.com/documentation/ihi0022/latest/) (not AXI-Stream!), mentions the following in Section A3.1 Clock and Reset:

> On Manager and Subordinate interfaces, there must be no combinatorial paths between input and output signals.

However, I checked the [AXI-Stream specification](https://developer.arm.com/documentation/ihi0051/latest/) and it is true that it doesn't include this restriction. This sentence does not appear in Section 2.8 Clock and Reset, nor anywhere else in the document.

In this sense, AXI handshaking is slightly more strict that AXI-Stream handshaking (I learned this today, as I thought that AXI-Stream also forbade combinatorial paths). I think I've never seen an AXI-Stream consumer that makes TREADY depend combinatorially on TVALID. In any case, I think that for this RFC allowing `ready` to depend combinatorially on `valid` is a good idea. The RFC is mainly about component interoperability, and having a consumer that makes `ready` depend combinatorially on `valid` doesn't affect whether it can be interfaced with a producer or not. Having this combinatorial path might not be a good idea for timing or architectural reasons, but it doesn't affect the "contract" of the interface handshaking.

whitequark:
> I think I've never seen an AXI-Stream consumer that makes TREADY depend combinatorially on TVALID.

I've asked around and I got a number of responses indicating that interconnect which allocates buffer space when an input stream becomes ready may need to do it combinatorially. That seemed like a good reason to allow it.

stafverhaegen-chipflow:
> I'm curious about why asserting valid combinatorially is allowed in the following:
> 
> > The receiver may assert `ready` in response to the transmitter asserting `valid` (including combinatorially).

I don't think this even works. I think it may mean that for a always asserted ready signal valid may never be asserted either.

text/0061-minimal-streams.md
```diff
@@ -0,0 +1,275 @@
...
+
+Data is transferred via the stream according to the following rules:
+1. On each cycle where both `valid` and `ready` are asserted, a transfer is performed: the contents of `payload` is conveyed from the transmitter to the receiver.
+2. Once the transmitter asserts `valid`, it must not deassert it until a transfer is performed.
+3. As an exception to rule (2), when the transmitter is reset it must deassert `valid`, unless it ties `valid` to `Const(1)` in the interface object.
+4. Once the transmitter asserts `valid`, it must not change the contents of `payload` until a transfer is performed.
+5. The transmitter must not assert `valid` in response to the receiver asserting `ready` (combinatorially or not).
+6. Once the receiver asserts `ready` it may deassert it at any time.
+7. The receiver may assert `ready` in response to the transmitter asserting `valid` (including combinatorially).
```
MegabytePhreak, line 38: I found rules 5 and 7 a little difficult to parse, although I believe I understand the intent. Some possible alternate formulations:
5. The transmitter must not condition the assertion of `valid` on the state of `ready` from the receiver
7. The receiver may condition the state of `ready` on the state of `valid` (or the contents of `payload`)(including combinatorially) 
(In addition to the previously discussed case of buffer allocation, another case where  ready may depend on both valid and data would be a block implementing routing or demultiplexing based on a `tdest` type signal in the payload)

text/0061-minimal-streams.md
```diff
@@ -0,0 +1,275 @@
...
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+### A: Scope of this proposal
+[scope-of-this-proposal]: #scope-of-this-proposal
+
+Streams are one of the most highly coveted planned features. While we could choose to not provide them at all (as an external library implementation is possible), this will make Amaranth less competitive and perpetuate the status quo of every complex downstream project implementing their own, incompatible, version of streams.
+
+This module is described as "minimal" because it is highly limited in scope. This has the desirable qualities that it is easier to implement, document, and teach; that it is more widely compatible with existing code using streams; and that it is amenable to further gradual expansion.
+- The payload is an opaque (to the `amaranth.lib.stream` module) shape that is not given further meaning.
```
MegabytePhreak, line 99: I appreciate this decision. I was recently working on a codebase that tried to implement AXI-Stream using systemverilog interfaces (for synthesis) and found a number of bugs and data size issues caused by the interface containing unused fields. I think starting from the simplest possible channel is the right answer. (The discussion of litex streams below, also resonates)

text/0061-minimal-streams.md
```diff
@@ -0,0 +1,275 @@
...
+
+Option (C2) does have the advantage of splitting the payload into individual signals, which are more visible in waveform diagrams and on module boundaries in Verilog. However, it has a serious drawback, which is that `amaranth.lib.wiring.connect()` will not connect a single member to multiple members. This means that streams can no longer be directly connected to the three signals made available by the existing FIFOs in `amaranth.lib.fifo`, requiring adaptation.
+
+Option (C3) affords the same flexibility in the payload as option (2), without requiring stream-specific FIFOs; instead, an existing FIFO with its `[rw]_ack`, `[rw]_rdy`, `[rw]_data` signals can be connected to a stream endpoint provided that the same signals are put into an `amaranth.lib.wiring.PureInterface`, as long as the `[rw]_data` signal has the right width.
+
+### D: Omitted features
+[omitted-features]: #omitted-features
+
+This proposal intentionally omits the following features:
+1. packetization (often implemented using `first` and/or `last` control signals)
```
MegabytePhreak, line 152: I don't think it's wrong to exclude this, but I do think that it is an important feature to get to quickly. Many use cases need packetizetion, and if it's not provided may invent their own version.

MegabytePhreak (commented):
I think this lays good groundwork for a useful streams abstraction.

text/0061-minimal-streams.md
```diff
@@ -0,0 +1,275 @@
...
+
+### B: Always-valid and always-ready streams
+[always-valid-and-always-ready-streams]: #always-valid-and-always-ready-streams
+
+Although, in general, a transmitter may produce a payload only on some cycles (communicated by `valid`) and a receiver may accept a payload only on some cycles (communicated by `ready`), there are important cases where dynamic flow control on either endpoint is undesirable, such as:
+- A SerDes/transceiver block would typically produce a symbol and consume a symbol on each cycle, with no ability to pause transmission. (If the two systems connected with a SerDes link are clocked asynchronously, dummy--but significant for protocol purposes--symbols are inserted for clock tolerance compensation.)
```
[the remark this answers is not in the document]
whitequark: Clocking the entire RX side with the recovered clock isn't really a good idea if you have some kind of bidirectional negotiation going on, or power management, or any other circumstance where the RX clock stops or glitches. But I can drop the parenthetical.

text/0061-minimal-streams.md
whitequark: "Condition the assertion" and "condition the state" are more confusing to me (introduces a new term used nowhere else in the spec; also I don't generally recall reading a lot of text using this terminology).

text/0061-minimal-streams.md
whitequark: An issue that I discovered is that there's no consensus on even the structure of trivial packets. Some people want `first` only, some `last` only, some both `first` and `last`. And this is before going into interaction of packetization with other features.

daniestevez: Maybe the fact that AXI-Stream uses `last` is a strong case for using `last`. I also find it easier to deal with than `first`. If the protocol has `last`, then when you see it asserted, you usually must reset things for the start of the next packet, and you can do it sequentially. A `first`, instead, usually forces you to deal with this situation combinationally, or to stop processing for one cycle.

text/0061-minimal-streams.md
daniestevez: AXI-Stream uses this wording:

> A Transmitter is not permitted to wait until TREADY is asserted before asserting TVALID. Once TVALID is
asserted, it must remain asserted until the handshake occurs.
>
> A Receiver is permitted to wait for TVALID to be asserted before asserting TREADY. It is permitted that a
Receiver asserts and deasserts TREADY without TVALID being asserted.

Maybe "wait for" is clearer than "in response to" or "condition". The main reason why these conditions are in place is to avoid a deadlock if the producer waits for `ready` and the consumer waits for `valid`, so one of them is not supposed to wait potentially indefinitely.

daniestevez: And while thinking more about this, I realize that maybe the current wording of 5 is perhaps not what is wanted here. In AXI-Stream it would be perfectly fine for a producer to assert `valid` in response to seeing `ready` asserted, provided that the producer also has a way in which it is guaranteed to assert `valid` even if the consumer never asserts `ready`. Is this the rule that is intended here?

whitequark: > Is this the rule that is intended here?

The intent is to exclude (a) comb feedback of `valid` from `ready` (which results in not-well-formed programs) and (b) in general `valid` being dependent on `ready` (which results in deadlocks).

I agree that these two are separate and would have to think about it more.

text/0061-minimal-streams.md
whitequark: > Maybe the fact that AXI-Stream uses `last` is a strong case for using `last`.

Not really. It is important to remain *interoperable* to some extent with existing protocols, yes, but the fact that AXI-Stream did something is not in itself a motivation for us doing something the same way. There are plenty of decisions in AMBA we should absolutely not replicate.

> If the protocol has `last`, then when you see it asserted, you usually must reset things for the start of the next packet, and you can do it sequentially. A `first`, instead, usually forces you to deal with this situation combinationally, or to stop processing for one cycle.

Other people (@zyp) argued for allowing `first` without `last` for cases where you don't know if you have a last byte, e.g. COBS. This is what I'm referring to as a lack of consensus on what the protocol should be.

text/0061-minimal-streams.md
daniestevez: Take for example a producer that asserts `valid` according to these rules (although I'd never write something like this other than for the sake of the example) :

1. Whenever `valid` is low,  a count-down is started on the next cycle.
2. When the count-down reaches the end, the counter stops in this "end-of-countdown" state.
3. If `ready` is high, the counter is set to the "end-of-countdown" state on the next cycle.
4. `valid` is the combinational OR of `ready` and the "end-of-countdown" state.

I think this would be okay in AXI-Stream (now that I've learned that combinational paths between inputs and outputs are not forbidden as in AXI), so the questions is if this producer would be allowed in Amaranth streams.

whitequark: Having re-read the spec:

> A Transmitter is not permitted to wait until TREADY is asserted before asserting TVALID. Once TVALID is asserted, it must remain asserted until the handshake occurs.

and this paragraph:

>  In AXI-Stream it would be perfectly fine for a producer to assert valid in response to seeing ready asserted, provided that the producer also has a way in which it is guaranteed to assert valid even if the consumer never asserts ready. Is this the rule that is intended here?

it is completely unclear to me why what you say is accurate.

Moreover, the following step would still create a combinatorial loop if the receiver tied `ready` to `valid`, so the example you provide does not motivate allowing this functionality.

> 4\. `valid` is the combinational OR of `ready` and the "end-of-countdown" state.

daniestevez: My example doesn't wait until TREADY is asserted. It waits until an internal counter expires. But it also asserts TVALID earlier if TREADY becomes asserted. I don't see why the spec would forbid this, even though it's a rather convoluted example.

> Moreover, the following step would still create a combinatorial loop if the receiver tied `ready` to `valid`.

You're right. Then maybe the fact that the AXI-Stream spec doesn't mention that combinatorial paths between input and output are not allowed is a typo rather than an intended choice to make it different from AXI?

whitequark: > My example doesn't wait until TREADY is asserted. It waits until an internal counter expires. But it also asserts TVALID earlier if TREADY becomes asserted.

In my view that counts as "waiting until TREADY" because that's a form of feedback.



> Then maybe the fact that the AXI-Stream spec doesn't mention that combinatorial paths between input and output are not allowed is a typo rather than an intended choice to make it different from AXI?

You only need to prohibit combinatorial paths in one direction (which it does), not in both, to prevent logic loops.

daniestevez: In any case I think that your last version https://github.com/amaranth-lang/rfcs/commit/1c9218781c1b8fc6071e3e0dda457cae9c916032 forbids combinatorial loops clearly enough, so the things I'm wondering are more about AXI-Stream than about this RFC.

zyp: > I don't see why the spec would forbid this, …

The issue is that if you have a source that under some conditions waits for `ready` before setting `valid` and a sink that under some conditions waits for `valid` before setting `ready` and connect them to each other, you've got a potential for a deadlock when they both come into the condition where they're waiting for the other end to make the first move.

To eliminate the possibility of deadlock, one of these must be forbidden, and having `valid` depend on the state of `ready` is the less useful one.

whitequark: @zyp I think @daniestevez's example doesn't have a deadlock. In it, the transmitter always makes forward progress, just at different speed.

zyp: Ah, yeah. I can't come up with a practical use for code that resembles the example, but since it neither has potential for deadlocks nor combinatorial loops, I don't see a specific reason to disallow it.

> Any form of feedback from `ready` to `valid` is prohibited.

This wording is too strict, as *de*assertion of `valid` very much depends on feedback from `ready`.

whitequark: > Ah, yeah. I can't come up with a practical use for code that resembles the example, but since it neither has potential for deadlocks nor combinatorial loops, I don't see a specific reason to disallow it.

I think having streams be push-only makes understanding the systems they are a part of easier, which to me is significant here.

> 
> > Any form of feedback from `ready` to `valid` is prohibited.
> 
> This wording is too strict, as _de_assertion of `valid` very much depends on feedback from `ready`.

Fixed, thanks. Now it's just the posedge that needs to happen without feedback.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-25#1711386368-1711386652;) this RFC on the 2024-03-25 core subsystem meeting. The disposition was to **merge**.

PR #62, opened 2024-03-22
RFC 62: The `MemoryData` class.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/mem-data/text/0062-memory-data.md)

### DIFF
--- a/text/0062-memory-data.md
+++ b/text/0062-memory-data.md
@@ -1,0 +1,121 @@
+- Start Date: 2024-03-25
+- RFC PR: [amaranth-lang/rfcs#62](https://github.com/amaranth-lang/rfcs/pull/62)
+- Amaranth Issue: [amaranth-lang/amaranth#1241](https://github.com/amaranth-lang/amaranth/issues/1241)
+
+# The `MemoryData` class
+
+## Summary
+[summary]: #summary
+
+A new class, `amaranth.hdl.MemoryData`, is added to represent the data and identity of a memory. It is used to reference the memory in simulation.
+
+## Motivation
+[motivation]: #motivation
+
+It is commonly useful to access a memory in a simulation testbench without having to create a special port for it. This requires storing some kind of a reference to the memory on the elaboratables.
+
+Currently, the object used for this is of a private `MemoryIdentity` class. It is implicitly created by `lib.memory.Memory` constructor, and passed to the private `_MemorySim{Read|Write}` objects when `__getitem__` is called.  This has a few problems:
+
+- the `Memory` needs to be instantiated in the constructor of the containing elaboratable; combined with its mutability and the occasional need to defer memory port creation to `elaborate`, this results in elaboratables that break when elaborated more than once
+- `amaranth.sim` currently requires a gross hack to recognize `Memory` objects in `traces`
+- occasionally, it is useful to have `Signal`s that are not included in the design proper, but are used to communicate between simulator processes; it could be likewise useful with memories, but it cannot work with the current code (since the `MemoryIdentity` that the simulator gets doesn't have enough information to actually create backing storage for the memory)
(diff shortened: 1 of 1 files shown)


text/0062-memory-data.md
```diff
@@ -0,0 +1,107 @@
...
+
+The `MemoryData` object is accessible via a new read-only `data` property on `Memory`. The existing `shape`, `depth`, `init` properties become aliases for `data.shape`, `data.depth`, `data.init`.
+
+The memory-related methods on the simulator context object introduced by RFC 36 become:
+
+- `memory_read(memory: MemoryData, address: int) -> int | Any`:
+  - Reads a value from the memory at the given address, then converts it via the same rules as `get()` using memory shape
+- `memory_write(memory: MemoryData, address: int, value: int | Any, mask: int = None)`
+  - Converts the value via the same rules as `set()` using memory shape, then writes it to the memory at the given address. If `mask` is not `None`, one of the following must be true:
+    - `memory.shape` is `usigned(n)`: `mask` is an `n`-bit bit mask of which bits of the value should be written
```
whitequark, line 79:
```suggestion
    - `memory.shape` is `unsigned(n)`: `mask` is an `n`-bit bit mask of which bits of the value should be written
```

text/0062-memory-data.md
```diff
@@ -0,0 +1,107 @@
...
+
+  - either `data`, or all three of `shape`, `depth`, `init` need to be provided, but not both
+  - if `data` is provided, it is used directly, and stored
+  - if `shape`, `depth`, `init` (and possibly `name`) are provided, they are used to create a `MemoryData`, which is then stored
+
+The `MemoryData` object is accessible via a new read-only `data` property on `Memory`. The existing `shape`, `depth`, `init` properties become aliases for `data.shape`, `data.depth`, `data.init`.
+
+The memory-related methods on the simulator context object introduced by RFC 36 become:
+
+- `memory_read(memory: MemoryData, address: int) -> int | Any`:
```
whitequark, line 75: Bikeshed: should these be called `memory_get` and `memory_set`?

wanda-phi:
Updated to add `MemoryData.Row` and reuse `get`/`set`/`changed` instead of dedicated `memory_read`/`memory_write`.

text/0062-memory-data.md
```diff
@@ -0,0 +1,116 @@
...
+- (proposed by [amaranth-lang/rfcs#64](https://github.com/amaranth-lang/rfcs/pull/64)) `tick().sample()`
+
+If `MemoryData.shape` is a `ShapeCastable`, the values are converted using `ShapeCastable.const` / `ShapeCastable.from_bits`. Otherwise, they need to be integers.
+
+`MemoryData` and `MemoryData.Row` instances can be added to the `traces` argument when writing a VCD file with the simulator.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+The ability to do a masked write introduced by RFC 36 is removed. It could be reintroduced (by adding `mask` to `set`, or by somehow accepting `sim.set(mem_data[row][sli:ce], ...)`), but it adds complexity and is redundant with `get` / modify / `set`.
```
whitequark, line 96: >  is redundant with `get` / modify / `set`.

Unfortunately this is not true. Several `add_process` processes writing the same signal will race if they all do an RMW, but not race if they do a masked write. (The same would *seem* to not be relevant for signals because you can write only a slice of a signal, but is actually broken in the same way because of our implementation.)

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-03-25#1711386712-1711388600;) this RFC on the 2024-03-25 core subsystem meeting. The disposition was to **merge**.

PR #63, opened 2024-03-22
Add an RFC for `lib.coding` removal.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/remove-lib-coding/text/0063-remove-lib-coding.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-04-08#1712595845-1712595964;) this RFC on the 2024-04-08 core subsystem meeting. The disposition was to **merge**.

PR #64, opened 2024-03-23
Amend RFC #36 with a concrete concurrency model

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/rfc-36-update/text/0036-async-testbench-functions.md)

### DIFF
--- a/text/0036-async-testbench-functions.md
+++ b/text/0036-async-testbench-functions.md
@@ -25,34 +25,22 @@ As an example, let's consider a simple stream interface with `valid`, `ready` an
 We can then implement `stream_send()` and `stream_recv()` functions like this:
 
 ```python
-@testbench_helper
 async def stream_recv(sim, stream):
-    await sim.set(stream.ready, 1)
-    await sim.tick().until(stream.valid)
-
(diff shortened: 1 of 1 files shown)


whitequark:
Notwithstanding that there are no dependencies between any of these items, I have implemented most of the proposed concurrency model in https://github.com/amaranth-lang/amaranth/pull/1231 and confirmed that it made it easier to reason about testbench concurrency.

whitequark:
Note: I pushed some last-minute changes to `until`/`repeat` which were queued but not previously published (sorry about that!). In short, these methods now always advance the timeline. This is because it does not take away any expressive power (a single `sim.get` before the method doesn't add significant overhead even with an optimized implementation offloading the waiting to C++) but in turn reduces the likelihood of observing a glitch, no matter the context. Also, the stream use case is actually better served by the new API.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-04-08#1712596005-1712596978;) this RFC on the 2024-04-08 core subsystem meeting. The disposition was to **merge**.

PR #65, opened 2024-04-05
Add an RFC for special formatting for structures and enums

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/format-struct-enum/text/0065-format-struct-enum.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-04-08#1712597056-1712599030;) this RFC on the 2024-04-08 core subsystem meeting. The disposition was to **merge**.

PR #66, opened 2024-04-16
RFC #66: Simulation time

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/simulation-time/text/0066-simulation-time.md).

### DIFF
--- a/text/0066-simulation-time.md
+++ b/text/0066-simulation-time.md
@@ -38,7 +38,8 @@ It is immutable and has a constructor accepting at most one named argument, givi
 - `Period(MHz: numbers.Real)`
 - `Period(GHz: numbers.Real)`
   - The argument will be scaled according to its SI prefix and its reciprocal used to calculate the closest integer femtosecond representation.
-    A value of zero will raise `ZeroDivisionError`.
+  - A value of zero will raise `ZeroDivisionError`.
+  - A negative value will raise `ValueError`.
 
 To convert it back to a number, the following properties are available:
 - `.seconds -> float`
@@ -54,6 +55,7 @@ To calculate the reciprocal frequency, the following properties are available:
 - `.megahertz -> float`
 - `.gigahertz -> float`
   - Accessing these properties when the period is zero will raise `ZeroDivisionError`.
+  - Accessing these properties when the period is negative will raise `ValueError`.
 
 The following operators are defined:
 - `.__lt__(other: Period) -> bool`
@@ -135,7 +137,7 @@ None.
 ## Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-- Should we disallow passing a negative frequency or accessing a frequency property on a negative period?
+None.
 
 ## Future possibilities
 [future-possibilities]: #future-possibilities
--- a/text/0066-simulation-time.md
+++ b/text/0066-simulation-time.md
@@ -25,17 +25,17 @@ A new type `Period` is introduced.
 It is immutable and has a constructor accepting at most one named argument, giving the following valid forms of construction:
 - `Period()`
   - Constructs a zero period.
-- `Period(seconds: numbers.Real)`
-- `Period(milliseconds: numbers.Real)`
-- `Period(microseconds: numbers.Real)`
-- `Period(nanoseconds: numbers.Real)`
-- `Period(picoseconds: numbers.Real)`
-- `Period(femtoseconds: numbers.Real)`
+- `Period(s: numbers.Real)`
+- `Period(ms: numbers.Real)`
+- `Period(us: numbers.Real)`
+- `Period(ns: numbers.Real)`
+- `Period(ps: numbers.Real)`
+- `Period(fs: numbers.Real)`
   - The argument will be scaled according to its SI prefix and used to calculate the closest integer femtosecond representation.
-- `Period(hertz: numbers.Real)`
-- `Period(kilohertz: numbers.Real)`
-- `Period(megahertz: numbers.Real)`
-- `Period(gigahertz: numbers.Real)`
+- `Period(Hz: numbers.Real)`
+- `Period(kHz: numbers.Real)`
+- `Period(MHz: numbers.Real)`
+- `Period(GHz: numbers.Real)`
   - The argument will be scaled according to its SI prefix and its reciprocal used to calculate the closest integer femtosecond representation.
     A value of zero will raise `ZeroDivisionError`.
 
@@ -136,12 +136,7 @@ None.
 
 - With the scope of `Period` extended beyond simulation, where do we put it?
 
-- Should we use shorter argument/property names like `MHz` instead of (or an alias in addition to) `megahertz`?
-
 - Should we disallow passing a negative frequency or accessing a frequency property on a negative period?
-
-- We could allow passing multiple constructor arguments and add them together -- should we?
-  - Probably not. Limited usefulness for durations and nonsensical for frequencies.
 
 ## Future possibilities
 [future-possibilities]: #future-possibilities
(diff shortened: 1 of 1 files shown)


text/0066-simulation-time.md
```diff
@@ -0,0 +1,119 @@
...
+- `.ns(duration: numbers.Real)`
+- `.ps(duration: numbers.Real)`
+- `.fs(duration: numbers.Real)`
+- `.Hz(frequency: numbers.Real)`
+- `.kHz(frequency: numbers.Real)`
+- `.MHz(frequency: numbers.Real)`
+- `.GHz(frequency: numbers.Real)`
+  - The argument will be scaled according to the SI prefix on the method and the `duration` or reciprocal of `frequency` then used to calculate the closest integer femtosecond representation.
+
+To convert it back to a number, the following properties are available:
```
lekcyjna123, line 38: Can we provide a convenience wrappers over time getters, to request the period rounded to int? If we specify time, then usually we know which unit is needed to get the exact int value. In such cases there will be always needed explicit cast which will produce boilerplate.

So I would like to propose to add:
- `.iseconds` -> int
- `.imiliseconds` -> int
- and so on

(of course the above function names are some example, which should probably by changed for something better)

text/0066-simulation-time.md
```diff
@@ -0,0 +1,119 @@
...
+
+Internally simulation time is an integer number of femtoseconds, but exposing that directly is not ergonomic and making it a `float` of seconds sacrifices precision the longer a simulation runs.
+
+Also, to quote @whitequark: Time is not a number.
+
+## Guide- and reference-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A new type `amaranth.sim.Period` is introduced.
+It is immutable, has no public constructor and is instead constructed through the following classmethods:
```
lekcyjna123, line 25: I am not sure if classmethod constructors should have such short names and the time getters the longer one. Usually the construction is more seldom than accessing. So maybe we can have constructors something like `.make_s(duration)`, `.make_ms(duration)`... and the access functions as `.get_s()`, `.get_ms()`,...

lekcyjna123 (commented):
For me it looks good. I left some minor bikeshedding comments.

text/0066-simulation-time.md
whitequark: Not in this case though. You will almost never use accessors. You should keep that in mind when evaluating the API design.

text/0066-simulation-time.md
whitequark: I don't see that as valuable due to how rare reading simulation time is expected to be.

whitequark:
Something that I realized is that this class would also be really convenient for specifying frequencies of clocks.

`amaranth.time`?

text/0066-simulation-time.md
```diff
@@ -0,0 +1,132 @@
...
+  - Multiplying a time period with or dividing it by a real number scales the time period.
+  - Dividing time periods gives the ratio between them.
+  - Modulo of time periods is the remainder and thus still a time period.
+    Potentially useful to calculate the phase offset between a timestamp and a clock period.
+  - Reflected operators that only accept `Period` operands are redundant and omitted.
+
+## Prior art
+[prior-art]: #prior-art
+
+None.
```
whitequark, line 117: I think CXXRTL using time as an 128-bit number of femtoseconds is prior art.

wanda-phi:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-06-10#1718039070-1718041320;) this RFC on the 2024-06-10 weekly meeting. Several concerns were raised about the RFC:

- (@whitequark) while it is clear what `time()` should return within a process or testbench, the semantics of `Simulator.time()` is unclear while the simulator is not active; the issues are:
  - when stopped after `advance()`, should the returned time be the time of the last processed event, or the time of the next event to be processed (the latter of which can be undefined if there are no more pending events)?
  - (@wanda-phi) when stopped after `run_until()`, it is somewhat surprising that `.time()` can (and often will) return something different than the requested time point.
  It has been proposed to remove `.time()` from `Simulator` entirely, keeping it only on `SimulatorContext`, where it is well-defined.
- (@whitequark, @zyp) should we have reciprocal properties on `Period` to recover a frequency from it?
  - it is not well-defined when the period is zero; however, we can just raise `ZeroDivisionError` in that case

Additionally, a few bikeshed discussions have happened:

- (@whitequark) the naming of `Simulator` and `SimulatorContext` method that returns elapsed time; the proposed alternatives are: `current_time()`, `next_time()`, `elapsed_time()`;
  - everyone seems reasonably happy with `elapsed_time()`?
- (@zyp) `hertz`, `kilohertz`, ... have been proposed for the frequency property accessors

whitequark:
Denominating as the RFC hasn't been updated to address feedback in https://github.com/amaranth-lang/rfcs/pull/66#issuecomment-2159096137.

text/0066-simulation-time.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- With the scope of `Period` extended beyond simulation, where do we put it?
+
+- The usual bikeshedding.
+  - Case, e.g. `.MHz` vs `.mhz`.
+
+- Should `Period` also have properties to return the reciprocal frequency?
```
whitequark, line 136: These have been added, so you can remove them from unresolved questions.

text/0066-simulation-time.md
```diff
@@ -0,0 +1,141 @@
...
+
+None.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- With the scope of `Period` extended beyond simulation, where do we put it?
+
+- The usual bikeshedding.
+  - Case, e.g. `.MHz` vs `.mhz`.
```
whitequark, line 134: I think we have discussed it before?

whitequark: (We did and there was no consensus.)

text/0066-simulation-time.md
whitequark: (It looks like there was also no clear consensus, so that's fine.)

zyp:
`datetime.timedelta` is actually decent prior art for what this RFC proposes; it serves a similar purpose and the set of supported operations is also pretty much identical. The difference is the lengths of the durations it's designed to work with.

Considering that `datetime.timedelta` only offers microsecond resolution and most clock periods tends to be an order of magnitude or two shorter, I don't think a conversion method would be very useful. Also, doing the conversion manually wouldn't be harder than `datetime.timedelta(microseconds = round(period.microseconds))`, which has the added benefit of being explicit about the precision loss.

whitequark:
Proposal: instead of `Period.kHz(1000)` do `Period(kHz=1000)`. This makes it very clear that it is a constructor and avoids any ambiguity. It's also the exact same number of characters.

zyp:
I've had the same thought, so if you also think that's a good idea, I can revise the RFC accordingly. This raises a few questions though:
- Do we allow argumentless `Period()` to construct a zero-period? (I'm leaning yes.)
- Do we allow multiple arguments that add up? E.g. `Period(s = 1, ms = 20) == Period(ms = 1020)`. (I'm leaning no, strong no for frequency arguments.)
- With properties and constructor arguments being separate namespaces, it'd be nice to keep them consistent (i.e. no `kHz` vs `kilohertz`), which raises the question of which to settle on? Alternatively we could do both as aliases.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-07-29#1722273084-1722273486;) this RFC on the 2024-07-09 weekly meeting. The disposition was to **merge**.

whitequark:
Thanks!

PR #67, opened 2024-05-29
Add RFC for combinable simulation triggers with `initial()`.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/sim-initial/text/0067-sim-initial.md)

### DIFF
--- a/text/0067-sim-initial.md
+++ b/text/0067-sim-initial.md
@@ -1,0 +1,120 @@
+- Start Date: 2024-06-03
+- RFC PR: [amaranth-lang/rfcs#67](https://github.com/amaranth-lang/rfcs/pull/67)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# Combinable simulation triggers with `initial()`
+
+## Summary
+[summary]: #summary
+
+A new method, `initial()`, is added to combinable triggers in the simulator. It makes the trigger fire once right after creation.
+
+## Motivation
+[motivation]: #motivation
+
+Consider the following process, meant to replace a bit of combinational logic:
+
+```
+a = Signal(16)
+b = Signal(16, init=1)
+o = Signal(32)
+
+async def process(ctx):
+    async for a_val, b_val in ctx.changed(a, b):
+        ctx.set(o, a_val * b_val)
+```
+
+It creates a trigger object that fires on every change of `a` and `b`, ensuring that `o = a * b`. However, if the initial state of `a`, `b`, and `o` is not consistent (eg. the first line is changed to `a = Signal(16, init=2)`), the trigger will not fire, and the process will fail to perform its function.
+
+It is not easily possible to fix this problem — `ctx.get()` cannot be used in a process and the only way to obtain signal values involves waiting on a trigger object. Further, even if `ctx.get()` was allowed in processes, solving the problem in the obvious way requires code duplication:
+
+```
+async def process(ctx):
+    ctx.set(o, ctx.get(a) * ctx.get(b))
+    async for a_val, b_val in ctx.changed(a, b):
+        ctx.set(o, a_val * b_val)
+```
+
+A similar problem also appears for flip-flops when asynchronous resets are involved. Consider the following code:
+
+```
+d = Signal()
+q = Signal()
+
+async def ff_process(ctx):
+    async for clk_edge, rst_val, d_val in ctx.tick().sample(d):
+        if rst_val:
+            ctx.set(q, 1)
+        elif clk_hit:
+            ctx.set(q, d_val)
+```
+
+While the code is meant to implement an asynchronous-set flip-flop, it fails to do so correctly when the `sync` domain has an asynchronous reset and the reset is initially asserted.
+
+
+## Guide-level and reference-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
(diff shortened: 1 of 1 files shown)


text/0067-sim-initial.md
```diff
@@ -0,0 +1,120 @@
...
+
+An increase in language complexity.
+
+The `.tick()` part is an obscure edge case.
+
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This RFC proposes to leave current semantics unchanged and add `.initial()` as something that has to be explicitly requested. Arguably, this can be seen as the wrong default. An alternative would be to make the "initial" behavior default whenever `async for` is involved. However, this in turn can be a very wrong default for other cases, and the flexibility of combinable triggers means that choosing a good default would require horrifying heuristics.
```
whitequark, line 102: Right. I think it's worth mentioning that we're discussing this because we coupled a performance improvement to the use of the `async for` syntax, which is why it's not viable to dedicate the `async for` syntax to processes.

text/0067-sim-initial.md
```diff
@@ -0,0 +1,120 @@
...
+
+The Verilog `always @*` construct has the same problem, resulting in pain and horrible workarounds whenever Verilog has to be emitted, including Amaranth's Verilog backend.
+
+The SystemVerilog `always_comb` construct implicitly performs an equivalent of `.initial()`.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- is `.tick().initial()` worth the complexity budget?
+- should we make writing the wrong implementation of combinational logic harder, somehow?
```
whitequark, line 115: Should we call it `.init()`, as a parallel to the `init=` argument, and unlike the Verilog construct?

text/0067-sim-initial.md
```diff
@@ -0,0 +1,120 @@
...
+The `.initial()` method is meant only for use with `async for` — passing the result of `.initial()` to plain `await` is an error.
+
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Both combinable and initial trigger objects are amended with one new method:
+
+- `initial()` 
+  - Create a new trigger object of the same type by copying the current object and appending an initial trigger. This causes the resulting trigger object to fire once immediately after first starting iteration by an `async for` loop. Awaiting such a trigger with plain `await` is an error. The `.initial()` trigger doesn't contribute any components to the yielded value.
```
whitequark, line 87: It is obvious what `.changed(...).initial()` would return. What about `.tick(...).initial()` though? Can it race with other processes at time 0?

whitequark:
My thoughts on the core of the RFC:

- Anyone who writes `async for _ in ctx.changed():` should be prepared to deal with glitches and make sure the body of the loop is either pure or at least idempotent, so it is probably safe to have `ctx.changed()` imply `.initial()`. I don't think we guarantee, or ever intend to guarantee, that `changed()` triggers are glitchless--even for testbenches!
- The specific narrow case of `async for _ in ctx.tick():` entering the loop initially if an async reset is asserted at the beginning of the simulation feels so narrowly scoped as to seem completely harmless; perhaps even more so than the former case. I cannot imagine any case at all where it would pose a problem--indeed, it's indistinguishable from the case of "all signals initialize to 0 and then a buffer added using a process sets an async reset to 1".

text/0067-sim-initial.md
wanda-phi: I don't know what kind of races you mean; all races should be prevented by other processes' changes not happening until the next delta cycle, right?

whitequark: It's a little more complicated than that. Normally, races (e.g. pulse on clk and rst simultaneously) are avoided because async signals don't toggle simultaneously with other signals. For example, clocks will be usually toggled by `add_clock` while everything else in the design is waiting for that tick, and resets will be usually asserted by testbenches as `ctx.set(rst, 1)` that doesn't change any other signal unless you opt into it via `ctx.set(Cat(...), ...)`.

However, things are weirder at time 0. Consider a situation where many combinational processes are used at time 0, all with `.initial()` or equivalent. These will all run in the same delta cycle, and may change a large amount of signals in that cycle. If several of those signals are a part of the same async control set you get a race.

In short, we don't actually *prevent* races, we make them harder to accidentally trigger via UI design, and in case of `.initial()` specifically the UI design part isn't there.

---

So knowing exactly what `.tick()` returns is important because the behavior during a race should be well-defined. Is it neither `clk` nor `rst? Is it `rst` but not `clk`? On careful reading I conclude that it's the latter but the RFC text should definitely say that explicitly.

whitequark:
I have an alternate proposal that achieves the same underlying goal (well-defined behavior at time 0):

1. `ctx.<any>.changed(...).<any>` acts as-if there was a change to an unspecified signal if it is awaited (via `await` or `async for`) at time 0;
2. `ctx.tick(domain).<any>` acts as-if there was an assertion of the asynchronous reset if `domain` has an asynchronous reset, the initial value of this asynchronous reset is 1, and it is awaited (via `await` or `async for`) at time 0.

The case (2) is niche enough that I can't imagine any change to the surface language pulling its weight in terms of complexity. The case (1) is less unambiguous, but since we don't guarantee absence of glitches, avoiding a very narrow sliver of spurious wakeups doesn't seem worth it.

wanda-phi:
abandoning in favor of an already-merged simpler fix (ie. [3d2cd15435bed62f026af64ba9c99b6f72759a97](https://github.com/amaranth-lang/amaranth/commit/3d2cd15435bed62f026af64ba9c99b6f72759a97))

PR #68, opened 2024-06-24
Add RFC for fine-grained output enable for I/O buffers

whitequark:
[Rendered](https://github.com/whitequark/amaranth-rfcs/blob/fine-grained-oe/text/0068-fine-grained-oe.md)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-07-01#1719855533-1719855990;) this RFC on the 2024-07-01 weekly meeting. The disposition was to **close**, since this RFC does not add any fundamentally new functionality, and nobody could state a compelling enough use case to justify the drawbacks.

PR #71, opened 2024-07-25
Add RFC for `EnumView.Matches`.

wanda-phi:
[Rendered](https://github.com/wanda-phi/rfcs/blob/enumview-matches/text/0071-enumview-matches.md).


whitequark:
Thanks!

PR #73, opened 2024-09-07
Add RFC for stricter connections.

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/stricter-connections/text/0073-stricter-connections.md).


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-09-09#1725901671-1725904693;) this RFC on the 2024-09-09 weekly meeting, and determined that it needs to cover `enum.View`/`enum.Const` as well as `data.Const`, and that the requirements on `data.View.eq` should take into account `Layout.cast(other.shape()`.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2024-09-16#1726499155-1726509308;) this RFC on the 2024-09-16 weekly meeting. The disposition was to **merge**, with the following changes:
- EnumView's `eq()` must accept any `ValueCastable` with an enumeration shape.

whitequark:
:tada: 

Thanks for the work, @zyp!

PR #74, opened 2025-01-22
Add RFC for structured VCD output

eigenform:
[Rendered](https://github.com/eigenform/rfcs/blob/main/text/0074-structured-vcd.md)


whitequark:
Thank you, this is very well written! My comments:

- Some examples of how the UI of GTKWave and Surfer will look like afterwards would be very helpful. (I think there are already RFCs which include media to use as an example.)
- An alternative is to keep the ability to use both formats, via the use of an environment variable (which is useful if you want to upgrade a third party codebase without modifying it, in possibly many places) and/or a parameter to `write_vcd()`.

text/0074-structured-vcd.md
```diff
@@ -204,9 +207,13 @@ $upscope $end
 - Should we continue to include the "flattened" [pure bit-vector] representation of aggregate signals in the VCD?
 - Does this feature need to be gated/opt-in by default?
 
+- The simulator currently depends on [westerndigitalcorporation/pyvcd](https://github.com/westerndigitalcorporation/pyvcd) when writing VCD files.
+  However, `pyvcd` only emits scope types defined by the VCD specification (and does not include the `vhdl_record`/`vhdl_array` scopes).
+  When implementing this RFC, should we continue relying on `pyvcd`, or does this warrant the addition of our own code in the simulator for writing VCD files?
```
whitequark, line 212: One reason to use pyvcd (VCD itself is not hard to emit) is its support for `.gtkw` files. I really don't want to have more GTKWave specific code in Amaranth...

whitequark:
Thinking about it, I feel that a prototype for this functionality would be not too difficult to build (you could depend on a git repo with a fork of pyvcd) and it would allow everyone impacted by the change to evaluate how it'll affect their workflow.

One consumer with little support for advanced VCD features that we need to care about is the Amaranth playground.

whitequark:
I closed the PR accidentally (Shift+Enter), apologies!

whitequark:
We have [discussed] this RFC on the 2025-05-12 meeting. The dispositon was to **merge**. The unresolved questions were approached as follows:
1. We should continue to have a "flattened" representation because, if nothing else, `data.Layout` permits having bits that are not a part of any field. Such bits would be completely invisible otherwise.
2. We should make this feature opt-out, both with a `write_vcd` argument and an environment variable. We will not port it to 0.5.x series.
3. We should work with pyvcd upstream to add these types; it is not reasonably expected to be a problem. (Pyvcd is a non-exposed implementation detail that isn't, in principle, a part of the RFC in first place.)
4. The Amaranth playground already has issues with structured data. It uses a library d3-wave that has "custom row renderers" that can be used to extend it in an application using it. We should update the playground code to emit custom row rendereds for structures (actually, we should've done it a while ago since right now strings don't display properly; amaranth-lang/playground#2).

The next steps would be:
1. To update the RFC text with the information above so it can be merged. (this is on @eigenform)
2. To merge it and open a tracking issue for it in the main repo. (this is on me)
3. To implement and ship it. (this is on @eigenform or anyone else interested)

[discussed]: https://libera.catirclogs.org/amaranth-lang/2025-05-12#1747069938-1747071431;

text/0074-structured-vcd.md
```diff
@@ -112,17 +112,34 @@ These images demonstrate the difference in GTKWave output:
...
 - The `module` scope defines a group of signals belonging to the same `Module`
 - The `vhdl_array` scope defines a group of signals belonging to an array (such as a signal with `ArrayLayout`)
-- The `vhdl_record` scope defines a structured group of signals (such as a signal with `StructLayout`)
+- The `vhdl_record` scope defines a structured group of signals (such as a signal with `StructLayout`, `UnionLayout`, or `FlexibleLayout`)
+
+After these changes, this functionality will be enabled by default in the VCD writer.
+These changes will **not** be backported to Amaranth 0.5.x.
+Users that require compatibility with the pre-RFC behavior are expected to either:
+
+- Opt-out on a case-by-case basis by passing arguments to `write_vcd()` (ie. `structured=False`)
```
whitequark, line 128: I suspect we may be using other VCD extensions later (in fact, we are using at least one already: strings; this is what's crashing the Playground currently).

How about `pure=True` and `AMARANTH_PURE_VCD=1`, to reflect the part that we deviate from the official VCD definition in IEEE 1800?

eigenform: Ooh, that's a good point - sounds great.

whitequark:
Tracking issue opened: https://github.com/amaranth-lang/amaranth/issues/1599. Can you add that to the body please?

whitequark:
Thank you for writing and updating the RFC!

PR #75, opened 2025-01-31
add RFC for property labels

rroohhh:
[Rendered](https://github.com/rroohhh/amaranth-rfcs/blob/property_label/text/0075-property-label.md)

### DIFF
--- a/text/0075-property-label.md
+++ b/text/0075-property-label.md
@@ -1,0 +1,75 @@
+- Start Date: (fill in with date at which the RFC is merged, YYYY-MM-DD)
+- RFC PR: [amaranth-lang/rfcs#0075](https://github.com/amaranth-lang/rfcs/pull/75)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# Property labels
+
+## Summary
+[summary]: #summary
+
+Add a `label` argument to the `Assume`, `Assert` and `Cover` statements, that gets converted to a statement label in the verilog backend.
+
+## Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


whitequark:
If the goal is to produce human friendly error messages, why are these tools not using the message field? I would say we should probably not add back `name` but figure out why is `message` not used.

wanda-phi:
> If the goal is to produce human friendly error messages, why are these tools not using the message field? I would say we should probably not add back `name` but figure out why is `message` not used.

SystemVerilog assertions do not *have* messages in the first place; relying on messages thus limits us to yosys-based tooling

on the other hand, the example given in the RFC could definitely be fixed on `sby` side

whitequark:
> SystemVerilog assertions do not _have_ messages in the first place; relying on messages thus limits us to yosys-based tooling

From my memory of talking to Jannis, they sort of do, in that `assert x else $error` is treated specially, or something along those lines.

rroohhh:
I care more about Jasper than `sby` at them moment to be honest. 
Jasper names the properties at elaboration time and `$error` can include formatting, so I guess that is why its not used?

whitequark:
@jix, would you happen to have advice here? I've never used Jasper.

jix:
I don't have any good advice regarding tool compatibility, I've also never used Jasper. I can confirm that using labels to identify properties seems what most FV users I've interacted with rely on, but I've also seen `else $error` being used in multiple places, but IIRC always in combination with also giving a label. I have no idea which tools the instances of that were originally targeting, though.

The SV spec defines the omission of an `else ...` "action block" to result in a call to `$error` if the assertion fails, so in that sense manually calling `$error` from the "action block" should be expected to work as intended, since it's defined to be the same thing (modulo the exact message used). With half of the spec only making sense for simulation, and non-simulation tools widely deviating from the spec even where they could follow it (whether for good or for questionable reasons) that all does end up being somewhat meaningless in practice though.

PR #76, opened 2025-05-10
RFC #76: `amaranth-boards` versioning policy.

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/amaranth-boards-versioning-policy/text/0076-amaranth-boards-versioning-policy.md).

### DIFF
--- a/text/0076-amaranth-boards-versioning-policy.md
+++ b/text/0076-amaranth-boards-versioning-policy.md
@@ -1,0 +1,49 @@
+- Start Date: (fill in with date at which the RFC is merged, YYYY-MM-DD)
+- RFC PR: [amaranth-lang/rfcs#76](https://github.com/amaranth-lang/rfcs/pull/76)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# `amaranth-boards` versioning policy
+
+## Summary
+[summary]: #summary
+
+Add a versioning policy for `amaranth-boards`.
+
+## Motivation
+[motivation]: #motivation
+
+Packages with direct (e.g. git) dependencies aren't allowed on PyPI, so by not publishing releases on PyPI, we're blocking downstream projects depending on `amaranth-boards` from publishing their releases on PyPi without having to do workarounds.
+
+## Explanation
+[guide-level-explanation]: #guide-level-explanation
+
+- Use a `0.x.y` versioning scheme.
+- Increment `y` and do a new release any time there's been done significant additions.
+- Increment `x`, reset `y` and do a new release any time there are breaking changes.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+None.
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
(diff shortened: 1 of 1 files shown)


whitequark:
The increment of `.y`, as well as releases, can be done automatically; I'd have to modify the existing build scripts we use a little but it's definitely feasible.

whitequark:
We have [discussed] this RFC on the 2025-05-12 meeting. The disposition was to **merge**, with the following comments:
1. This should be implemented by automatically releasing every commit to `main` whose HEAD message doesn't contain `[breaking change]` or `[breaking-change]` with a version like `v0.x.y` where `y` is the number of commits since the last breaking change, and automatically tagging every commit that does with a tag `v0.(x+1).0`.
2. A breaking change is anything that requires manual intervention for code that previously worked fine. I.e. a change to a board file which previously had two pins incorrectly swapped would not be a breaking change; a change to names of resources across multiple boards would be.

[discussed]: https://libera.catirclogs.org/amaranth-lang/2025-05-12#1747071559-1747072293;

whitequark:
Thank you!

fischermoseley:
I'd like to take a shot at implementing this! I've got a PyPI package that depends on the board definitions, and it'd be very convenient for users to not need to install it separately. 

However, I was hoping to get clarification as to how the `y` in `0.x.y` should be updated. The RFC seems to specify two different behaviors - one simply increments `y`, and the other counts the number of commits since the last breaking change. In principle these should be equivalent, but what if a PR containing more than one commit is merged? If we were counting commits, we'd release with non-consecutive version numbers - which may be confusing to users.

We could just require commit squashing on `main`, but wanted to ask if it'd be acceptable to just increment `y` as that may be more robust.

whitequark:
> I'd like to take a shot at implementing this! I've got a PyPI package that depends on the board definitions, and it'd be very convenient for users to not need to install it separately.

This would be very much appreciated, and I'll do my best to support your efforts. (I'm currently applying to become a permanent resident in the UK, which saps some of the time I have for OSS...)

> In principle these should be equivalent, but what if a PR containing more than one commit is merged? If we were counting commits, we'd release with non-consecutive version numbers - which may be confusing to users.
> 
> We could just require commit squashing on `main`, but wanted to ask if it'd be acceptable to just increment `y` as that may be more robust.

My thoughts are that counting commits is very easy and already something we do in other subprojects, but counting merged PRs (without actually introducing merge commits) is tricky and I'm not sure how we'd approach this.

Non-consecutive version numbers are a little confusing, yes, but given that Amaranth already uses this technique with the "local suffix" and Yosys does the same it doesn't feel profoundly out of line.

Ultimately I'm fine with either implementation (and the RFC is indeed ambiguous as written), but I'd like to have the versioning code to be robust so counting commits seems preferable from that point of view.

fischermoseley:
Thanks for the clarification! I didn't realize that this inherited from a precedent elsewhere, and I've gone ahead and put together a first pass in [#263](https://github.com/amaranth-lang/amaranth-boards/pull/263) on the `amaranth-boards` repo.

Best of luck with the paperwork! I'm definitely not in a hurry to get this merged, and happy to help however I can.

PR #77, opened 2025-05-10
RFC #77: Stream port name conventions.

zyp:
[Rendered](https://github.com/zyp/amaranth-rfcs/blob/stream-port-name-conventions/text/0077-stream-port-name-conventions.md).

### DIFF
--- a/text/0077-stream-port-name-conventions.md
+++ b/text/0077-stream-port-name-conventions.md
@@ -1,0 +1,61 @@
+- Start Date: (fill in with date at which the RFC is merged, YYYY-MM-DD)
+- RFC PR: [amaranth-lang/rfcs#77](https://github.com/amaranth-lang/rfcs/pull/77)
+- Amaranth Issue: [amaranth-lang/amaranth#0000](https://github.com/amaranth-lang/amaranth/issues/0000)
+
+# Stream port name conventions
+
+## Summary
+[summary]: #summary
+
+Settle on and document name conventions for stream components.
+
+## Motivation
+[motivation]: #motivation
+
+Having a name convention for stream ports makes stream components easier to work with and allows automating the connection of pipelines like this:
+
+```
+pipeline = [component_a, component_b, component_c]
+for upstream, downstream in itertools.pairwise(pipeline):
+    wiring.connect(m, upstream.output, downstream.input)
+```
+
+## Guide- and reference-level explanation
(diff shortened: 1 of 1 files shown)


whitequark:
> Which exact names should we settle on?
> - i / o

These are the ones I've been using recently.

> - input / output

These are the ones I've been using in the past. I settled on `i` and `o` because if you need to add more streams, expanding them to `i_foo` and `o_bar` feels logical, while `input_foo` and `output_bar` too verbose and somewhat at odds with how we've been approaching core design (see e.g. `lib.io`).

> - w_stream / r_stream

I would say these, as well as `r` and `w`, are acceptable when there is *reading and writing* involved (like from/to a memory), but not in general. I'm undecided on whether I want a `stream.FIFO` (or whatever we call it when we land it) should use `r`/`w` or `i`/`o`. I've been using `r`/`w` in Glasgow, but then I found myself mentally mapping them to `i`/`o` anyway, at which point standardizing might make more sense.

> - sink / source

This is the one option I'm decidedly against. The names are [very confusing](https://github.com/amaranth-lang/amaranth/issues/317#issuecomment-899407394) with plenty of in-the-field feedback to that end from newcomers to LiteX. They don't shorten well (`si`/`so` perhaps, but then we're back to `i`/`o`). And they have no relation to the existing language/stdlib at all, making them an additional burden to learn. If you've used `lib.wiring` for a while you're probably used to how inputs become outputs when flipped, and know that `flipped(self.i)` is an output; now you have to learn that `flipped(self.sink)` is a source.

whitequark:
We have [discussed](https://libera.catirclogs.org/amaranth-lang/2025-06-09#1749488435-1749490007;) this RFC on the 2025-06-09 weekly meeting. We have decided to **merge** this RFC, while resolving the unresolved questions as follows:
- Whenever the name of a stream member of the component interface includes an indication of its direction, the indication must match one of the following forms:
  - For `In` members, either the complete name is `i` (for components that have one primary input), or the name must start with `i_` (for more complex components).
  - For `Out` members, either the complete name is `o` (for components that have one primary input), or the name must start with `o_` (for more complex components).
- It is also acceptable to have a name that doesn't indicate directionality explicitly, e.g. `requests` or `packets`.

whitequark:
@zyp Do you think you could update the PR and open a tracking issue?

PR #79, opened 2025-05-30
Add an RFC for `PortLike.with_direction`.

wanda-phi:


whitequark:
Thank you for submitting the RFC!

